PR #2, opened 2022-07-01
Implementation of TopoJSON objects

IIITM-Jay:
This PR deals with the implementation to parse a file (Topojson objects) into OCaml representation according to the specifications being mentioned [here](https://github.com/topojson/topojson-specification)

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -9,7 +9,7 @@
 
 (license MIT)
 
-(authors "Patrick Ferris")
+(authors "Patrick Ferris", "Jay Dev Jha")
 
 (maintainers "pf341@patricoferris.com")
 
--- a/dune-project
+++ b/dune-project
@@ -9,9 +9,9 @@
 
 (license MIT)
 
-(authors "Patrick Ferris", "Jay Dev Jha")
+(authors "Patrick Ferris" "Jay Dev Jha")
 
-(maintainers "pf341@patricoferris.com")
+(maintainers "patrick@sirref.org")
 
 (package
  (name topojson)
@@ -20,9 +20,11 @@
   This library provides a functor for building a module for manipulating and parsing Topojson into OCaml.")
  (depends
    (brr :with-test)
+   (bos :with-test)
    (bechamel-notty (and (>= 0.1.0) :with-test))
    (alcotest :with-test)
-   (ezjsonm  :with-test)))
+   (ezjsonm  :with-test)
+   (geojson  :with-test)))
 
 (package
  (name topojsonm)
--- a/geojson.opam.template
+++ b/geojson.opam.template
@@ -1,0 +1,3 @@
+pin-depends: [
+  ["ocaml-geojson" "git+https://github.com/geocaml/ocaml-geojson"]
+]
--- a/geojson.opam.template
+++ b/geojson.opam.template
@@ -1,3 +1,0 @@
-pin-depends: [
-  ["ocaml-geojson" "git+https://github.com/geocaml/ocaml-geojson"]
-]
--- a/src/topojson/dune
+++ b/src/topojson/dune
@@ -1,3 +1,4 @@
 (library
  (public_name topojson)
- (name topojson))
+ (name topojson)
+ (libraries geojson))
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -19,6 +19,12 @@ module Make (J : Intf.Json) = struct
 
   module Geometry = struct
     type json = J.t
+    type properties = [ `None | `Null | `Obj of (string * json) list ]
+
+    let properties_or_null = function
+      | `None -> []
+      | `Null -> [ ("properties", J.null) ]
+      | `Obj v -> [ ("properties", J.obj v) ]
 
     let keys_in_use =
       [
@@ -32,7 +38,7 @@ module Make (J : Intf.Json) = struct
         "geometries";
       ]
 
-    let foreign_members json =
+    let foreign_members_of_json json =
       match J.to_obj json with
       | Ok assoc ->
           List.filter (fun (k, _v) -> not (List.mem k keys_in_use)) assoc
@@ -110,13 +116,12 @@ module Make (J : Intf.Json) = struct
       let parse_coords coords = J.to_array (decode_or_err J.to_float) coords
       let base_of_json json = parse_by_type json parse_coords typ
 
-      let to_json ?bbox ?properties ?(foreign_members = []) position =
+      let to_json ?bbox ?(properties = `None) ?(foreign_members = []) position =
         J.obj
           ([
-             ("type", J.string typ);
-             ("coordinates", Position.to_json position);
-             ("properties", Option.(value ~default:J.null properties));
+             ("type", J.string typ); ("coordinates", Position.to_json position);
            ]
+          @ properties_or_null properties
           @ bbox_to_json_or_empty bbox
           @ foreign_members)
     end
@@ -134,13 +139,14 @@ module Make (J : Intf.Json) = struct
 
       let base_of_json json = parse_by_type json parse_coords typ
 
-      let to_json ?bbox ?properties ?(foreign_members = []) positions =
+      let to_json ?bbox ?(properties = `None) ?(foreign_members = []) positions
+          =
         J.obj
           ([
              ("type", J.string typ);
              ("coordinates", J.array Position.to_json positions);
-             ("properties", Option.(value ~default:J.null properties));
            ]
+          @ properties_or_null properties
           @ bbox_to_json_or_empty bbox
           @ foreign_members)
     end
@@ -153,13 +159,10 @@ module Make (J : Intf.Json) = struct
       let parse_arcs arcs = J.to_array (decode_or_err J.to_int) arcs
       let base_of_json json = parse_by_arcs json parse_arcs typ
 
-      let to_json ?bbox ?properties ?(foreign_members = []) arc =
+      let to_json ?bbox ?(properties = `None) ?(foreign_members = []) arc =
         J.obj
-          ([
-             ("type", J.string typ);
-             ("arcs", Arcs.to_json arc);
-             ("properties", Option.(value ~default:J.null properties));
-           ]
+          ([ ("type", J.string typ); ("arcs", Arcs.to_json arc) ]
+          @ properties_or_null properties
           @ bbox_to_json_or_empty bbox
           @ foreign_members)
     end
@@ -176,13 +179,10 @@ module Make (J : Intf.Json) = struct
 
       let base_of_json json = parse_by_arcs json parse_arcs typ
 
-      let to_json ?bbox ?properties ?(foreign_members = []) arcs =
+      let to_json ?bbox ?(properties = `None) ?(foreign_members = []) arcs =
         J.obj
-          ([
-             ("type", J.string typ);
-             ("arcs", J.array Arcs.to_json arcs);
-             ("properties", Option.(value ~default:J.null properties));
-           ]
+          ([ ("type", J.string typ); ("arcs", J.array Arcs.to_json arcs) ]
+          @ properties_or_null properties
           @ bbox_to_json_or_empty bbox
           @ foreign_members)
     end
@@ -203,13 +203,10 @@ module Make (J : Intf.Json) = struct
 
       let base_of_json json = parse_by_arcs json parse_arcs typ
 
-      let to_json ?bbox ?properties ?(foreign_members = []) arcs =
+      let to_json ?bbox ?(properties = `None) ?(foreign_members = []) arcs =
         J.obj
-          ([
-             ("type", J.string typ);
-             ("arcs", J.array (J.array J.int) arcs);
-             ("properties", Option.(value ~default:J.null properties));
-           ]
+          ([ ("type", J.string typ); ("arcs", J.array (J.array J.int) arcs) ]
+          @ properties_or_null properties
           @ bbox_to_json_or_empty bbox
           @ foreign_members)
     end
@@ -227,13 +224,13 @@ module Make (J : Intf.Json) = struct
 
       let base_of_json json = parse_by_arcs json parse_arcs typ
 
-      let to_json ?bbox ?properties ?(foreign_members = []) arcs =
+      let to_json ?bbox ?(properties = `None) ?(foreign_members = []) arcs =
         J.obj
           ([
              ("type", J.string typ);
              ("arcs", J.array (J.array (J.array J.int)) arcs);
-             ("properties", Option.(value ~default:J.null properties));
            ]
+          @ properties_or_null properties
           @ bbox_to_json_or_empty bbox
           @ foreign_members)
     end
@@ -249,20 +246,27 @@ module Make (J : Intf.Json) = struct
 
     and t = {
       geometry : geometry;
-      properties : ((string * json) list) option;
+      properties : properties;
       foreign_members : (string * json) list;
     }
 
-    let g ?properties ?(foreign_members = []) geo =
+    let v ?(properties = `None) ?(foreign_members = []) geo =
       { geometry = geo; properties; foreign_members }
 
     let geometry t = t.geometry
     let properties t = t.properties
+    let foreign_members t = t.foreign_members
+
+    let properties_of_json json =
+      match J.find json [ "properties" ] with
+      | Some j -> if J.is_null j then `Null else `Obj (decode_or_err J.to_obj j)
+      | None -> `None
 
     let rec base_of_json json =
-      let fm = foreign_members json in
-      match (J.find json [ "type" ], J.find json [ "properties" ]) with
-      | Some typ, properties -> (
+      let fm = foreign_members_of_json json in
+      let properties = properties_of_json json in
+      match J.find json [ "type" ] with
+      | Some typ -> (
           match J.to_string typ with
           | Ok "Point" ->
               Result.map (fun g ->
@@ -277,7 +281,7 @@ module Make (J : Intf.Json) = struct
                   { geometry = LineString g; properties; foreign_members = fm })
               @@ LineString.base_of_json json
           | Ok "MultiLineString" ->
-              Result.map (fun v ->
+              Result.map (fun g ->
                   {
                     geometry = MultiLineString g;
                     properties;
@@ -315,10 +319,10 @@ module Make (J : Intf.Json) = struct
                        geometries"))
           | Ok typ -> Error (`Msg ("Unknown type of geometry " ^ typ))
           | Error _ as e -> e)
-      | None, _ ->
+      | None ->
           Error
             (`Msg
-              "A Geojson text should contain one object with a member `type`.")
+              "A TopoJSON text should contain one object with a member `type`.")
 
     let rec to_json ?bbox t =
       match t.geometry with
@@ -346,12 +350,9 @@ module Make (J : Intf.Json) = struct
                ("type", J.string "GeometryCollection");
                ("geometries", J.list to_json c);
              ]
+            @ properties_or_null t.properties
             @ bbox_to_json_or_empty bbox
-            @ foreign_members
-            @ geo_properties properties)
-
-    let foreign_members (_, fm) = fm
-    let geo_properties (_, props) = props
+            @ t.foreign_members)
   end
 
   module Topology = struct
@@ -375,7 +376,7 @@ module Make (J : Intf.Json) = struct
         "geometries";
       ]
 
-    let foreign_members json =
+    let foreign_members_of_json json =
       match J.to_obj json with
       | Ok assoc ->
           List.filter (fun (k, _v) -> not (List.mem k keys_in_use)) assoc
@@ -397,7 +398,7 @@ module Make (J : Intf.Json) = struct
                     (decode_or_err (J.to_array (decode_or_err J.to_float)))))
               arcs
           in
-          let fm = foreign_members json in
+          let fm = foreign_members_of_json json in
           Ok { objects = geometries; arcs; foreign_members = fm }
       | _, _ -> Error (`Msg "No objects and/or arcs field in Topology object!")
 
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -43,7 +43,7 @@ module Make (J : Intf.Json) = struct
           | t when t = typ -> p_c coords
           | t -> Error (`Msg ("Expected type of `" ^ typ ^ "' but got " ^ t)))
 
-    let parse_by_type json p_a typ =
+    let parse_by_arcs json p_a typ =
       match (J.find json [ "type" ], J.find json [ "arcs" ]) with
       | None, _ ->
           Error
@@ -138,7 +138,7 @@ module Make (J : Intf.Json) = struct
       let typ = "LineString"
       let v arc = arc
       let parse_arcs arcs = J.to_array (decode_or_err J.to_int) arcs
-      let base_of_json json = parse_by_type json parse_arcs typ
+      let base_of_json json = parse_by_arcs json parse_arcs typ
 
       let to_json ?bbox ?(foreign_members = []) arc =
         J.obj
@@ -157,7 +157,7 @@ module Make (J : Intf.Json) = struct
         try J.to_array (decode_or_err LineString.parse_arcs) arcs
         with Failure m -> Error (`Msg m)
 
-      let base_of_json json = parse_by_type json parse_arcs typ
+      let base_of_json json = parse_by_arcs json parse_arcs typ
 
       let to_json ?bbox ?(foreign_members = []) arcs =
         J.obj
@@ -180,7 +180,7 @@ module Make (J : Intf.Json) = struct
           J.to_array (decode_or_err (J.to_array (decode_or_err J.to_int))) arcs
         with Failure m -> Error (`Msg m)
 
-      let base_of_json json = parse_by_type json parse_arcs typ
+      let base_of_json json = parse_by_arcs json parse_arcs typ
 
       let to_json ?bbox ?(foreign_members = []) arcs =
         J.obj
@@ -200,7 +200,7 @@ module Make (J : Intf.Json) = struct
         try J.to_array (decode_or_err Polygon.parse_arcs) arcs
         with Failure m -> Error (`Msg m)
 
-      let base_of_json json = parse_by_type json parse_arcs typ
+      let base_of_json json = parse_by_arcs json parse_arcs typ
 
       let to_json ?bbox ?(foreign_members = []) arcs =
         J.obj
@@ -297,7 +297,7 @@ module Make (J : Intf.Json) = struct
       | Error _ -> []
 
     type t = {
-      objects : (string * json) list;
+      objects : (string * Geometry.t) list;
       arcs : Geometry.Position.t array array;
     }
 
@@ -305,6 +305,9 @@ module Make (J : Intf.Json) = struct
       match (J.find json [ "objects" ], J.find json [ "arcs" ]) with
       | Some objects, Some arcs ->
           let* objects = J.to_obj objects in
+          let geometries =
+            List.map (fun (k, v) -> (k, Geometry.base_of_json v)) objects
+          in
           let* arcs =
             J.to_array
               (decode_or_err
@@ -322,7 +325,8 @@ module Make (J : Intf.Json) = struct
     match J.find json [ "type" ] with
     | Some typ -> (
         match J.to_string typ with
-        | Ok "Topology" -> Topology.of_json json
+        | Ok "Topology" ->
+            Ok (Topology (Result.get_ok @@ Topology.of_json json))
         | Ok s -> Error (`Msg ("Expected `Topology` but got " ^ s))
         | Error _ as e -> e)
     | None -> Error (`Msg "Could not find Topology type")
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -38,6 +38,9 @@ module Make (J : Intf.Json) = struct
           List.filter (fun (k, _v) -> not (List.mem k keys_in_use)) assoc
       | Error _ -> []
 
+    let geo_properties properties =
+      if properties = [] then [] else [ ("properties", J.obj properties) ]
+
     let parse_by_type json p_c typ =
       match (J.find json [ "type" ], J.find json [ "coordinates" ]) with
       | None, _ ->
@@ -117,8 +120,7 @@ module Make (J : Intf.Json) = struct
            ]
           @ bbox_to_json_or_empty bbox
           @ foreign_members
-          @ if properties = [] then [] else [ ("properties", J.obj properties) ]
-          )
+          @ geo_properties properties)
     end
 
     module MultiPoint = struct
@@ -142,8 +144,7 @@ module Make (J : Intf.Json) = struct
            ]
           @ bbox_to_json_or_empty bbox
           @ foreign_members
-          @ if properties = [] then [] else [ ("properties", J.obj properties) ]
-          )
+          @ geo_properties properties)
     end
 
     module LineString = struct
@@ -159,8 +160,7 @@ module Make (J : Intf.Json) = struct
           ([ ("type", J.string typ); ("arcs", Arcs.to_json arc) ]
           @ bbox_to_json_or_empty bbox
           @ foreign_members
-          @ if properties = [] then [] else [ ("properties", J.obj properties) ]
-          )
+          @ geo_properties properties)
     end
 
     module MultiLineString = struct
@@ -180,8 +180,7 @@ module Make (J : Intf.Json) = struct
           ([ ("type", J.string typ); ("arcs", J.array Arcs.to_json arcs) ]
           @ bbox_to_json_or_empty bbox
           @ foreign_members
-          @ if properties = [] then [] else [ ("properties", J.obj properties) ]
-          )
+          @ geo_properties properties)
     end
 
     module Polygon = struct
@@ -205,8 +204,7 @@ module Make (J : Intf.Json) = struct
           ([ ("type", J.string typ); ("arcs", J.array (J.array J.int) arcs) ]
           @ bbox_to_json_or_empty bbox
           @ foreign_members
-          @ if properties = [] then [] else [ ("properties", J.obj properties) ]
-          )
+          @ geo_properties properties)
     end
 
     module MultiPolygon = struct
@@ -230,8 +228,7 @@ module Make (J : Intf.Json) = struct
            ]
           @ bbox_to_json_or_empty bbox
           @ foreign_members
-          @ if properties = [] then [] else [ ("properties", J.obj properties) ]
-          )
+          @ geo_properties properties)
     end
 
     type geometry =
@@ -253,11 +250,10 @@ module Make (J : Intf.Json) = struct
       { geometry = geo; properties; foreign_members }
 
     let geometry t = t.geometry
-    let properties t = t.properties
 
     let rec base_of_json json =
       let fm = foreign_members json in
-      let props = properties t.properties in
+      let props = geo_properties properties in
       match J.find json [ "type" ] with
       | Some typ -> (
           match J.to_string typ with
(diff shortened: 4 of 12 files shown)


IIITM-Jay:
Hi @patricoferris , actually I got stuck at finding the array index type. I am wondering is there a way to get the type  of the variable used in `OCaml` (I mean to say to check the data type whether a float or int) or need to create a function for that.
Need some guidance to proceed further...

dune-project
```diff
@@ -9,7 +9,7 @@
 
 (license MIT)
 
-(authors "Patrick Ferris")
+(authors "Patrick Ferris", "Jay Dev Jha")
```
patricoferris: Don't think it needs the comma :))

src/topojson/dune
```diff
@@ -1,3 +1,4 @@
 (library
  (public_name topojson)
- (name topojson))
+ (name topojson)
+ (libraries geojson))
```
patricoferris, line 4: We'll need to add this to the dependencies in the topojson package in the `dune-project` file.

src/topojson/topojson.ml
```diff
@@ -1 +1,16 @@
 (* Implentation of TopoJSON Objects *)
+
+module Geometry = struct
+  module LineString = struct
```
patricoferris: This is heading in the right direction. I've spent a bit of time having a closer look at the specification and I'm not convinced anymore that we will be able to share that much with `Geojson` because the underlying geometries change quite a bit (and therefore so do the functions for getting things like `foreign_members`). In particular, for `LineString` it can only contain "arcs" now and not "coordinates". So looks like we'll have to reimplement a few bits copying from Geojson.

This is the general structure I was thinking after reading the spec. What do you think?

```ocaml
module Make (G : Geojson.S) = struct
  module Geometry = struct

    module Point = G.Geometry.Point
    module LineString = struct
      type t = int array
      let v arcs = arcs
    end

    type geometry =
      | Point of Point.t
      | LineString of LineString.t
    
    and t = { geometry : geometry; id : G.json option; properties : (string * G.json) list; foreign_members : (string * G.json) list; }
  end

  module Topology = struct
    type t
  end

  type t = Topology of Topology.t | Geometry of Geometry.t
end
```

Note I've functorised over some implementation of `Geojson`. I'm also copying a little bit from https://github.com/georust/topojson.

IIITM-Jay:
Hi Patrick:)), A very Good Evening!
I have pushed some commits, requesting you to have a review and guidance on further implementations...
Is I am doing in a correct way?

I am thinking to work further on completing the rest of the geometry objects implementation and the topology objects. Once this is done, I am thinking to work upon the transformation and quantization...

IIITM-Jay:
Hi Patrick:)), A very Good Evening!
I have pushed some commits, requesting you to have a review and guidance on further implementations...
Is I am doing in a correct way?

I am thinking to work further on completing the rest of the geometry objects implementation and the topology objects. Once this is done, I am thinking to work upon the transformation and quantization...

src/topojson/topojson.ml
```diff
@@ -1 +1,182 @@
...
+    module MultiLineString = struct
+      type t = LineString.t array
+
+      let typ = "MultiLineString"
+      let v arcs = arcs
+
+      let to_json ?arcs ?bbox ?(foreign_members = []) arcs =
+        J.obj
+          ([ ("type", J.string typ); ("arcs", J.array (J.array J.int) arcs) ]
+          (* @ arcs *)
```
```suggestion
```

Guessing all of these can be removed :))

src/topojson/topojson.ml
```diff
@@ -1 +1,182 @@
...
+      | MultiPoint of MultiPoint.t
+      | LineString of LineString.t
+      | MultiLineString of MultiLineString.t
+    (* | Polygon of Polygon.t
+       | MultiPolygon of MultiPolygon.t
+       | Collection of t list *)
+
+    and t = {
+      geometry : geometry;
+      arcs : int array;
```
patricoferris: Things in `type t` must be shared across all geometries for example having a `properties` field. Not all geometries have `arcs` so I think that can be removed here.

```suggestion
```

src/topojson/topojson.ml
```diff
@@ -1 +1,182 @@
...
+      id : G.json option;
+      properties : (string * G.json) list;
+      foreign_members : (string * G.json) list;
+    }
+  end
+
+  module Topology = struct
+    type t = { objects : (string * G.json) list; arcs : int array array }
```
```suggestion
    type t = { objects : (string * G.json) list; arcs : Position.t array array }
```

From the spec:

> A topology must have an “arcs” member whose value is an array of arrays of positions

src/topojson/topojson.ml
```diff
@@ -1 +1,182 @@
...
+
+    let of_json json =
+      match (J.find json [ "objects" ], J.find json [ "arcs" ]) with
+      | Some objects, arcs ->
+          Result.map (fun v -> { objects; arcs }) (J.to_obj v)
```
patricoferris: This isn't quite right, and doesn't currently compile:

Something like this might be better (if you define the appropriate `let+` and `let*` or feel free to use `Result.map` and `Result.bind` if you prefer).

```ocaml
 let of_json json =
      match (J.find json [ "objects" ], J.find json [ "arcs" ]) with
      | Some objects, Some arcs -> (
        let* objects = J.to_obj objects in
        let+ arcs = J.to_array (decode_or_err (J.to_array (decode_or_err J.to_int))) arcs in
        { objects; arcs }
      )
      | _, _ -> Error (`Msg "No objects and/or arcs field in Topology object!")
````

**Review by patricoferris — COMMENTED** [review]
This is heading in the right direction. Let's focus this PR on just parsing a file into the OCaml representation and worry about the transformation and quantisation after.

**Comment by patricoferris** [comment]
I've also enabled OCaml-CI for this repo :))

File: src/topojson/topojson.ml (hunk 4 above)
Comment (IIITM-Jay) [reply src/topojson/topojson.ml]: Agreed :+1:  :))

File: src/topojson/topojson.ml (hunk 5 above)
Comment (IIITM-Jay) [reply src/topojson/topojson.ml]: Yup @patricoferris , I firstly thought to not to consider here, but just wanted to confirm from you side. it should be inside the particular geometries that requires so :+1: 1:

File: src/topojson/topojson.ml (hunk 6 above)
Comment (IIITM-Jay) [reply src/topojson/topojson.ml]: ohh! yeah , I  was just missing this point as it somewhat confused with 

> **A topology must have a member with the name “arcs” whose value is an array of arcs.**

but when I saw the examples out there, it indeed implies the same i.e., ```positions``` only

File: src/topojson/topojson.ml (hunk 7 above)
Comment (IIITM-Jay) [reply src/topojson/topojson.ml]: This seems somewhat newer to me the ```let*``` & ```let+``` things will explore more on this...
Thanks for letting me know these Patrick:))

**Comment by IIITM-Jay** [comment]
Hi @patricoferris , 
I have one confusion, **is the  `arcs` type _same_** throughout the json file **_or different_** referring to the below examples?
```
{
          "type": "LineString",
          "properties": {
            "prop0": "value0",
            "prop1": 0
          },
          "arcs": [0]
        },
```
Here, it seems as an **_`int array`_** type

```
{
          "type": "Polygon",
          "properties": {
            "prop0": "value0",
            "prop1": {
              "this": "that"
            }
          },
          "arcs": [[-2]]
        }
```
Whereas, it seems to be of type **_`int array array`_**

**Comment by IIITM-Jay** [comment]
And, like as you said, the arcs here,
```
"arcs": [
    [[4000, 0], [1999, 9999], [2000, -9999], [2000, 9999]],
    [[0, 0], [0, 9999], [2000, 0], [0, -9999], [-2000, 0]]
  ]
```
is of type **_`Position.t array array`_**

So, I think we need to declare the `arc` type specifically in each geometry module bcz don't think will work with global declarartion

**Comment by patricoferris** [comment]
The `"arcs"` will be different depending on the geometry:

> For type “LineString”, the “arcs” member must be an array of arc indexes
> For type “MultiLineString”, the “arcs” member must be an array of LineString arc indexes

So if for `LineString` it is an array of arc indexes i.e. `int array` then for a `MultiLineString` an array of such arcs would be an `int array array`

**Comment by IIITM-Jay** [comment]
yeah, thought so only before writing this comment but then thought to just confirm once , thanks Patrick:))

**Comment by IIITM-Jay** [comment]
Hi Patrick:)),
Have the following doubts regarding the latest commit:

1. Throws the following error , I don't understand that why it shows like that, as `Position.t` is itself of type `float array` quite confusing :disappointed: 
```
 File "src/topojson/topojson.ml", line 337, characters 21-25:
337 |           { objects; arcs }
                           ^^^^
Error: This expression has type float array array
       but an expression was expected of type Position.t array array
       Type float is not compatible with type Position.t = float array 
```
2. Secondly, How to access the **`Position`** module form **`Geometry`** module into the **`Topology`** module ? I copy pasted that module into the Topology module as I tried simply using **`Geometry.Position.t`** but didn't work

_Requesting for your valuable suggestions and guidance on the same._

File: src/topojson/topojson.ml [hunk 8]
```diff
@@ -1 +1,351 @@
...
+    let of_json json =
+      match (J.find json [ "objects" ], J.find json [ "arcs" ]) with
+      | Some objects, Some arcs ->
+          let* objects = J.to_obj objects in
+          let* arcs =
+            J.to_array
+              (decode_or_err (J.to_array (decode_or_err J.to_float)))
+              arcs
+          in
+          { objects; arcs }
```
```suggestion
          Ok { objects; arcs }
```

If your final let-statement isn't a `let+` then you'll need to wrap this in an `Ok`.

src/topojson/topojson.ml
```diff
@@ -1 +1,351 @@
...
+
+    type t = { objects : (string * json) list; arcs : Position.t array array }
+
+    let of_json json =
+      match (J.find json [ "objects" ], J.find json [ "arcs" ]) with
+      | Some objects, Some arcs ->
+          let* objects = J.to_obj objects in
+          let* arcs =
+            J.to_array
+              (decode_or_err (J.to_array (decode_or_err J.to_float)))
```
patricoferris: As you pointed out the error message for arcs is

```ocaml
Error: This expression has type float array array
       but an expression was expected of type Position.t array array
       Type float is not compatible with type Position.t = float array
```

We have given arcs the type `Position.t array array` i.e. an array, of an array of positions. The compiler helpfully tells us that a `Position.t` is in fact a `float array` so if we substitute that in, arcs should be of type `float array array array` but your parsing function only makes a `float array array`. It needs an extra `J.to_array` I think :))

src/topojson/topojson.ml
```diff
@@ -1 +1,351 @@
...
+
+    let keys_in_use = [ "type"; "arcs"; "objects"; "transform"; "bbox" ]
+
+    let foreign_members json =
+      match J.to_obj json with
+      | Ok assoc ->
+          List.filter (fun (k, _v) -> not (List.mem k keys_in_use)) assoc
+      | Error _ -> []
+
+    module Position = struct
```
patricoferris: You don't need to copy this

patricoferris:
> Secondly, How to access the Position module form Geometry module into the Topology module ? I copy pasted that module into the Topology module as I tried simply using Geometry.Position.t but didn't work

This compiles for me, what was the error you got when trying?

```ocaml
type t = { objects : (string * json) list; arcs : Geometry.Position.t array array }
```

src/topojson/topojson.ml
IIITM-Jay: Yeah, got it now... :+1:

IIITM-Jay:
Yeah now working, actually the issue was: (in the ```_intf.ml``` file)
```
184 |     arcs : Geometry.Position.t array array;
                 ^^^^^^^^^^^^^^^^^^^
Error: Unbound module Geometry

```
but now resolved...

IIITM-Jay:
Hi Patrick:))
There are some issues like:
_1. Facing this issue, like what does it exactly implying :_
```
File "src/topojson/topojson.ml", line 238, characters 17-45:
238 |               @@ LineString.base_of_json json
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: This expression has type (int array, [ `Msg of string ]) result
       but an expression was expected of type (LineString.t, 'a) result
       Type int array is not compatible with type LineString.t = Arcs.t array 
       Type int is not compatible with type Arcs.t = int array 
```

**2. In the test file, how to bind the arc labels, got stuck** 
```
File "test/topojson/test.ml", line 47, characters 7-36:
47 |   | Ok { Topojson.Topology.objects } ->
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error (warning 9 [missing-record-field-pattern]): the following labels are not bound in this record pattern:
arcs
Either bind these labels explicitly or add '; _' to the pattern.
```

IIITM-Jay:
Also, the important issue:
```
File "src/topojson/topojson.ml", line 1:
Error: The implementation src/topojson/topojson.ml
       does not match the interface src/topojson/.topojson.objs/byte/topojson.cmi:
        ... ... ... In module Make.Topology:
       The module `Geometry' is required but not provided
       File "src/topojson/topojson_intf.ml", line 182, characters 2-50:
         Expected declaration
```

IIITM-Jay:
If I use `;_`               or   

     ` | Ok { Topojson.Topology.objects; arcs } ->`

for solving **`Error (warning 9 [missing-record-field-pattern]):`**, I am getting like this:

```
File "test/topojson/test.ml", line 47, characters 7-42:
47 |   | Ok { Topojson.Topology.objects; arcs } ->
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: This pattern matches values of type Topojson.Topology.t
       but a pattern was expected which matches values of type Topojson.t

```

patricoferris:
For 

```
File "src/topojson/topojson.ml", line 1:
Error: The implementation src/topojson/topojson.ml
       does not match the interface src/topojson/.topojson.objs/byte/topojson.cmi:
        ... ... ... In module Make.Topology:
       The module `Geometry' is required but not provided
       File "src/topojson/topojson_intf.ml", line 182, characters 2-50:
         Expected declaration
```

If you follow to where the compiler is warning, it is that you have declared in the interface for `Topojson` that the `Topology` module should contain a `Geometry` module:

https://github.com/IIITM-Jay/ocaml-topojson/blob/d399baf191ed3d6e8f161d098b5a177c40d06870/src/topojson/topojson_intf.ml#L182

In the implementation no such module exists. I can see why you have done this in order to be able to reference `Geometry.Position.t`. I recommend following the ocaml-geojson way of achieving this by moving the `Topology` module type into `S`:

```ocaml
module type S = sig
  type json
  (** The internal representation of JSON *)

  module Geometry : Geometry with type json = json
  module Topology : sig
    type t = {
      objects : (string * json) list;
      arcs : Geometry.Position.t array array;
    }
  end

...
```

patricoferris:
For 

```
File "test/topojson/test.ml", line 47, characters 7-42:
47 |   | Ok { Topojson.Topology.objects; arcs } ->
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: This pattern matches values of type Topojson.Topology.t
       but a pattern was expected which matches values of type Topojson.t
```
This is because you have defined a `Topojson.t` to be:

```ocaml
type t = Topology of Topology.t | Geometry of Geometry.t
```

So you would need to pattern-match, something like:

```ocaml
 | Ok (Topology { Topojson.Topology.objects; arcs } ) ->
```

However, I don't think we need to provide those two different type `t` (it's called a variant in OCaml, or sum type). Because the spec iiuc says:

> TopoJSON always consists of a single topology object. A topology may contain any number of named geometry objects. 

So type t can be simplified to:

```ocaml
  type t = Topology.t
```

IIITM-Jay:
The reason to define `Topojson.t` as
`type t = Topology of Topology.t | Geometry of Geometry.t`

is that the value of the  `type` of TopoJSON object can be any of the `Geometry types `  or the ` Topology ` and not only the `Topology` (I think so)

> The value of the type member must be one of: “Topology”, “Point”, “MultiPoint”, “LineString”, “MultiLineString”, “Polygon”, “MultiPolygon” or “GeometryCollection”. The case of the type member values must be as shown here.

what do you think?  🤔

patricoferris:
Good point! In which case it should stay as is, you just need to add the `Ok (Topology...)` :))

IIITM-Jay:
To get the following issue resolved:
```
File "src/topojson/topojson.ml", line 318, characters 15-22:
318 |           Ok { objects; arcs }
                     ^^^^^^^
Error: This expression has type (string * json) list
       but an expression was expected of type (string * Geometry.t) list
       Type json is not compatible with type
         Geometry.t = Geometry.geometry * (string * json) list 
```

We need to make some changes to the line:
**`          let* objects = J.to_obj objects in`**

I am not sure, but thinking so. but got stuck as wondering if _**`J.to_obj `**_  is indeed required ?
I think the `objects` is getting the type as `(string * json) list` due to  `to_obj objects` only, I think... 
bcoz we have already declare  the `objects` type as `(string * Geometry.t)` list everywhere, how can we get through this

geojson.opam.template
```diff
@@ -0,0 +1,3 @@
+pin-depends: [
+  ["ocaml-geojson" "git+https://github.com/geocaml/ocaml-geojson"]
```
patricoferris: The package name is `geojson` not `ocaml-geojson`. I think it also requires a version, so you can use `.dev`

patricoferris (commented):
Yes, we can first convert it to the raw objects association list (`string * json list`) and then map over it converting the `json` to a `Geometry.t`. That's what this line was hinting at: 

https://github.com/geocaml/ocaml-topojson/commit/f5e9dd8db93d2cedd06bb7be3ec7c87a2cd219b0#diff-3b476409d0d0f36229c018b403b265af01138c5dd8f720f585232d7b2eae2988R308

IIITM-Jay:
Looks like we have already done the following:

> Yes, we can first convert it to the raw objects association list (string * json list) and then map over it converting the json to a Geometry.t. 

but still the issue persists:

```
318 |           Ok { objects; arcs }
                     ^^^^^^^
Error: This expression has type (string * json) list
       but an expression was expected of type (string * Geometry.t) list
       Type json is not compatible with type
         Geometry.t = Geometry.geometry * (string * json) list 
```

patricoferris:
You need to replace it with 

```ocaml
Ok { objects = geometries; arcs }
```

or rename the variable to `objects`

IIITM-Jay:
Actually, I have tried both the ways before also, but was getting the same sort of errors, that's why asked so.
To solve that issue I have used:

```
type t = {
      objects : (string * (Geometry.t, [ `Msg of string ]) result) list;
```
so, now the compiler is not showing up that error.
But I am not sure whether I have done it in a correct way or still some better approach can be done for this
What do you think Patrick...

geojson.opam.template
patricoferris: geojson.dev

patricoferris: The file name needs to be `topojson.opam.template`

src/topojson/topojson.ml
```diff
@@ -315,19 +318,45 @@ module Make (J : Intf.Json) = struct
...
+
+    let to_json ?bbox { objects; arcs; foreign_members } =
+      J.obj
+        ([
+           ("type", J.string "Topology");
+           ("objects", J.obj objects);
+           ("arcs", J.array (J.array (J.array J.float)) arcs);
+         ]
+        @ bbox_to_json_or_empty bbox
+        @ foreign_members)
```
IIITM-Jay: Hi @patricoferris ,While working with this function , got some issues with  `("objects", J.obj objects);` as we already know the fact that:

> A topology must have a member with the name “objects” whose value is another object. The value of each member of this object is a geometry object.

so I think `J.obj` would be correct (or J.list have to be used, got confused), hence I tried with it.
`J.obj` has type `(string * json)` list but `objects` has type `(string * Geometry.t) list` so, we indeed wanted `(string * json)` list type in this and not particularly  `(string * Geometry.t) list`....

Got confused, how to do it?

IIITM-Jay: Have spent some time thinking about it, but not getting it exactly

patricoferris:
Just a pushed a commit to get the CI running -- opam's metadata is poorly documented, so thought that would be quicker. Be sure to pull that commit in :))

src/topojson/topojson.ml
patricoferris: It's essentially the inverse to what we did for `of_json` where we converted from `json` to `Geometry.t` and now we want to convert back: 

```ocaml
("objects", J.obj (List.map (fun (k, v) -> (k, Geometry.to_json v)) objects));
```

IIITM-Jay: Oh never thought the sol would be too easy, only the reverse technique has to be applied, Things seems a bit easier than I am thinking... learning so much.
Thanks Patrick:))

IIITM-Jay:
> Just a pushed a commit to get the CI running -- opam's metadata is poorly documented, so thought that would be quicker. 

Thanks Patrick, I saw the changes and got quite surprised with the fact that embedding the `sha`key of the successfully passed last commit of `geojson` would help to get the CI running. I too have observed some hash keys of some repo (like `ocaml/opam`)in the build details. Learning much more everyday :hugs: 

By _opam's metadata is poorly documented_, you mean the `dune-project` file, right?

And working on `test`, what I am missing in this one, can we do like this (showing errors, missing something :disappointed: ):

```
let () =
  let s = read_file "./test_cases/files/exemplar.json" in
  let json = Ezjsonm.value_from_string s in
  let topojson_obj = Topojson.of_json json in
  match topojson_obj with
  | Ok v -> (
      match Topojson.to_json v with
      | (Topology f,[]) -> (v,f)
      | _ -> assert false)
  | _ -> assert false
  | Error (`Msg m) -> failwith m
```

src/topojson/topojson_intf.ml
```diff
@@ -1 +1,230 @@
...
+      objects : (string * Geometry.t) list;
+      arcs : Geometry.Position.t array array;
+      foreign_members : (string * json) list;
+    }
+
+    include Json_conv with type t := t and type json := json
+  end
+
+  type topojson = Topology of Topology.t | Geometry of Geometry.t
+  type t
```
patricoferris, line 215: I think we want `type topojson` to be type `t` ?

test/topojson/test.ml
```diff
@@ -1 +1,54 @@
...
+module Topojson = Topojson.Make (Ezjsonm_parser)
+
+let () =
+  let s = read_file "./test_cases/files/exemplar.json" in
+  let json = Ezjsonm.value_from_string s in
+  let topojson_obj = Topojson.of_json json in
+  match topojson_obj with
+  | Ok v -> (
+      match Topojson.to_json v with
+      | (Topology f,[]) -> (v,f)
```
patricoferris: > what I am missing in this one, can we do like this (showing errors, missing something 😞 )

So like the previous error we're dealing with the typechecker again. We have in general two possibilities. Either the function is not returning the type we expected (so we need to change the function) or we are assuming the wrong type (and need to change the code where the error is). Here's the error:

```
File "test/topojson/test.ml", line 51, characters 8-23:
51 |       | (Topology f,[]) -> (v,f)
             ^^^^^^^^^^^^^^^
Error: This pattern matches values of type 'a * 'b
       but a pattern was expected which matches values of type Topojson.json
```

Here you have pattern-matched on a pair hence `'a * 'b`. Do we expect a pair? The actual type is `Topojson.json`, is this expected? You have first converted the `json` to a `t` with `Topojson.of_json json` and then you convert it back to the raw `json` with `Topojson.to_json v` so that's right. This means your pattern-match is incorrect.

src/topojson/topojson_intf.ml
```diff
@@ -1 +1,230 @@
...
+  and t = geometry * (string * json) list
+
+  val foreign_members : t -> (string * json) list
+  (** [foreign_members t] will extract name/value pair of a foreign member from
+      t (a topojson object) *)
+
+  include Json_conv with type t := t and type json := json
+end
+
+module type Topology = sig
```
patricoferris: Do we still need this, `Topology` is defined in `S`

IIITM-Jay:
Hi Patrick:)) yeah surely will follow up with your comments. Have left some messages on Zulip as well regarding feedback. Requesting you to kindly look into that.
Thanks again for giving your valuable time in reviewing. Seems busy yesterday !

src/topojson/topojson_intf.ml
IIITM-Jay: Yeah sure, we don't need this, just forgot to remove ...

src/topojson/topojson_intf.ml
IIITM-Jay: I have done this bcoz I think due to the `bbox` implementation:

```
type topojson = Topology of Topology.t | Geometry of Geometry.t
  and t = { topojson : topojson; bbox : float array option }

  let topojson_to_t tjson bbox = { topojson = tjson; bbox }
```

That's why kept like that, still wanted your suggestions on this...

test/topojson/test.ml
IIITM-Jay: Will work on it to successfully match the pattern

IIITM-Jay:
Hi Patrick, Good Morning:))
```
51 |       | Topojson.Topology f -> v
             ^^^^^^^^^^^^^^^^^^^
Error: This pattern matches values of type Topojson.topojson
       but a pattern was expected which matches values of type Topojson.json
```
Here, we indeed wanted the to match with the values of type **`Topojson.topojson`** and of course, we need to check the **`Topojson.to_json`**, which is being defined as:
 ```
let to_json = function
    | { topojson = Topology f; bbox } -> Topology.to_json ?bbox f
```
I have spent time trying to rectify it, but not getting the way exactly, need your valuable guidance on the same.

**Comment by IIITM-Jay** [comment]
and for this 
> I think we want `type topojson` to be type `t` ?

I have done this bcoz I think due to the bbox implementation:

```
type topojson = Topology of Topology.t | Geometry of Geometry.t
  and t = { topojson : topojson; bbox : float array option }
  let topojson_to_t tjson bbox = { topojson = tjson; bbox }
```
That's why kept like that, still wanted your suggestions on this...

**Comment by patricoferris** [comment]
> Here, we indeed wanted the to match with the values of type Topojson.topojson and of course, we need to check the Topojson.to_json, which is being defined as:

I think things are getting confused here. There are three types I think we are mixing up.

 - `type json` -- this is the OCaml representation of the _raw JSON_ i.e. something like `{ "hello": "world" }`.
 - `type topojson` -- is a topojson object either a Topology or a Geometry
 - `type t` -- is a topojson object but with the shared extra data (like the bounding box)

The function `val to_json` returns `json` (the raw JSON i.e. in this test `Ezjsonm.value`). `val of_json` produces a `type t` whose field member `topojson` is of type `topojson`, right now there is no way to access that because there is no function to do so and the type `t` is abstract.

I've fixed these issues which I think should clear things up in this commit https://github.com/geocaml/ocaml-topojson/commit/3631c539c7ce0e9642db7c28d7245934f2a27d77 -- I would recommend that you `git cherry-pick` that commit into this branch. The implementation needs fixed however as you can see from the failing test https://ci.ocamllabs.io/github/geocaml/ocaml-topojson/commit/3631c539c7ce0e9642db7c28d7245934f2a27d77/variant/debian-11-4.08_opam-2.1 (hint: I would have a look at your foreign members code ;)) 

Hopefully that's helpful, let me know if anything is not clear :))

**Comment by IIITM-Jay** [comment]
What an awesome explanation :+1: , Thank you so much Patrick for this :pray: , the main thing that I was missing:

>  there is no way to access that because there is no function to do so and the type t is abstract.

I think for the failing test, `properties` need to be implemented in every `Geometry` , I guess bcoz `properties` being eliminated as when we receive the `json` after running test. we need something to support the following since in the `exemplar.json` which we are testing contains `properties` field in every geometry types:

> A geometry object may have a member with the name “properties”. The value of the properties member is an object (any JSON object or a JSON null value)

needs your suggestions on the same...

**Comment by IIITM-Jay** [comment]
In `geojson` we have `properties` field in `Feature/ Feature Collection` only and not in any of the `Geometry type`.
What do you think...

**Comment by patricoferris** [comment]
Yes, this is another subtle difference between the two I think. A geometry in this case will be:

```ocaml
type geometry =
      | Point of Point.t
      | MultiPoint of MultiPoint.t
      | LineString of LineString.t
      | MultiLineString of MultiLineString.t
      | Polygon of Polygon.t
      | MultiPolygon of MultiPolygon.t
      | Collection of t list

    and t = {
      geometry: geometry;
      properties: (string * json) list;
      foreign_members: (string * json) list;
    }
```

And you will have to plumb through `properties` similarly to how `foreign_members` is.

Note as well, another reason the test is failing is because `geometries` is in the output JSON object twice

IIITM-Jay:
Added the `properties` field but that needs your guidance on the same.

> Note as well, another reason the test is failing is because geometries is in the output JSON object twice

What could be the reason for that, don't think it could be `foreign_members`... but not sure

IIITM-Jay:
For the error `Collection` type 
` | Collection of t list`

we had before
` t = geometry * (string * json) list`

but how to do now as we changed it to records

src/topojson/topojson.ml
```diff
@@ -263,25 +276,27 @@ module Make (J : Intf.Json) = struct
...
-          Point.to_json ?bbox ~foreign_members point
-      | MultiPoint mp, foreign_members ->
-          MultiPoint.to_json ?bbox ~foreign_members mp
-      | LineString ls, foreign_members ->
-          LineString.to_json ?bbox ~foreign_members ls
-      | MultiLineString mls, foreign_members ->
-          MultiLineString.to_json ?bbox ~foreign_members mls
-      | Polygon p, foreign_members -> Polygon.to_json ?bbox ~foreign_members p
-      | MultiPolygon mp, foreign_members ->
+      | Point point, foreign_members, properties ->
```
patricoferris: This is matching a triple of values, but we have (which is good I think) changed to represented the value as a record. So you will need to pull the values out of the record.

src/topojson/topojson.ml
```diff
@@ -222,34 +228,41 @@ module Make (J : Intf.Json) = struct
...
       match J.find json [ "type" ] with
       | Some typ -> (
           match J.to_string typ with
           | Ok "Point" ->
-              Result.map (fun v -> (Point v, fm)) @@ Point.base_of_json json
+              Result.map (fun v -> (Point v, fm, props))
+              @@ Point.base_of_json json
           | Ok "MultiPoint" ->
-              Result.map (fun v -> (MultiPoint v, fm))
+              Result.map (fun v -> (MultiPoint v, fm, props))
```
patricoferris: Same thing applies here, you are constructing a triple when you need to be constructing the record.

src/topojson/topojson.ml
```diff
@@ -203,14 +208,15 @@ module Make (J : Intf.Json) = struct
...
+      let to_json ?bbox ?(properties = []) ?(foreign_members = []) arcs =
         J.obj
           ([
              ("type", J.string typ);
              ("arcs", J.array (J.array (J.array J.int)) arcs);
            ]
           @ bbox_to_json_or_empty bbox
-          @ foreign_members)
+          @ foreign_members
+          @ properties)
```
patricoferris: These are slightly wrong. The properties are a value associated with the `"properties"` field so you need to construct it similarly to how `"type"` and `"arcs"` are. Something like:

```ocaml
 @ (if properties = [] then [] else [ ("properties", J.obj properties) ])
```

patricoferris:
> What could be the reason for that, don't think it could be foreign_members... but not sure

The foreign members implementation is not quite right for two reasons:

1. `geometries` needs to be in `keys_in_use` -- that's why it is appearing twice.
2. `keys_in_use` is applied for every geometry which is not right. For example it treats `arcs` as a key in use which is only correct for geometries that use that field. `arcs` is a foreign member for a `Point`. I can fix this later in a different PR.

IIITM-Jay:
Like the below one, we need to do...?

```
let v ?(properties = []) ?(foreign_members = []) geo =
      { geometry = geo; properties; foreign_members }

    let geometry t = t.geometry
    let properties t = t.properties

    let rec base_of_json json =
      let fm = foreign_members json in
      let props = properties t.properties in
      match J.find json [ "type" ] with
      | Some typ -> (
          match J.to_string typ with
          | Ok "Point" ->
              Result.map (fun v ->
                  {
                    geometry = Point v;
                    properties = props;
                    foreign_members = fm;
                  })
              @@ Point.base_of_json json
```

pushed the commit also,...

IIITM-Jay:
The problem is what needs to be written in       **`let props = properties t in`**

```
let rec base_of_json json =
      let fm = foreign_members json in
      let props = properties t.properties in
```

Also, 
> This is matching a triple of values, but we have (which is good I think) changed to represented the value as a record. So you will need to pull the values out of the record.

This I am not getting, how to do it for below...
```
let rec to_json ?bbox = function
      | Point point, foreign_members, properties ->
          Point.to_json ?bbox ~foreign_members ~properties point
```

patricoferris:
For 

```ocaml
let rec to_json ?bbox = function
      | Point point, foreign_members, properties ->
          Point.to_json ?bbox ~foreign_members ~properties point
```

It could be changed to something like 

```ocaml
let rec to_json ?bbox t = match t.geometry with 
      | Point point ->
          Point.to_json ?bbox ~foreign_members:t.foreign_members ~properties:t.properties point
```

And for 

```ocaml
 let rec base_of_json json =
      let fm = foreign_members json in
      let props = properties t.properties in
      match J.find json [ "type" ] with
      | Some typ -> (
          match J.to_string typ with
          | Ok "Point" ->
              Result.map (fun v ->
                  {
                    geometry = Point v;
                    properties = props;
                    foreign_members = fm;
                  })
              @@ Point.base_of_json json
```

Not quite, you need to actually get the `properties` JSON value (like we got the foreign members). Here you've tried to access the properties of the OCaml value for the geometry which doesn't make sense because we're actually building that value in this function.

IIITM-Jay:
**The last part not completely understood, pushed a commit also so that it will be easier to point out my mistakes ...** 

1. **_Like we should firstly create the function like this :_**
```
 let geo_properties properties =
      if properties = [] then [] else [ ("properties", J.obj properties) ]
 
```
2. **_Secondly, using it here as:_**
```
let to_json ?bbox ?(properties = []) ?(foreign_members = []) position =
        J.obj
          ([
             ("type", J.string typ); ("coordinates", Position.to_json position);
           ]
          @ bbox_to_json_or_empty bbox
          @ foreign_members
          @ geo_properties properties) 
```
3. **_(i) but now, how to use it here:_**
```
  let rec base_of_json json =
      let fm = foreign_members json in
      let props = geo_properties properties in
      match J.find json [ "type" ] with
   
```
**_(ii)And here while doing in the records, is the following way not correct after resolving of the above point 3(i):_**
 ```
Result.map (fun g ->
                  {
                    geometry = Point g;
                    properties = props;
                    foreign_members = fm;
                  })
              @@ Point.base_of_json json 
```

**Comment by patricoferris** [comment]
So I think there's some confusion again here. The `base_of_json` function takes raw JSON (in the form of the JSON parser so for example, `type json = Ezjsonm.value` and produces an OCaml value for that object. So:

```json
{
        "type": "Point",
        "properties": {
          "prop0": "value0"
        },
        "coordinates": [102, 0.5]
      },
```

Needs to produce the `Point` geometry along with extracting `properties` field to put in the record (note there are no foreign members here). You are trying to use the `geo_properties` function you defined, but that is only useful for going the other way around -- producing JSON for the OCaml value. 

You need to extract the properties (if there are any) in step 3 (i) from the JSON using functions like `J.find` and `J.to_obj` etc. Yes 3 (ii) is fine I think if you fix 3 (i) 👍

IIITM-Jay:
Hi @patricoferris ,
I have made some changes to the `properties` field as per my understanding. I have changed the properties type as 
`      properties : json option;` since I don't find any harm in doing so (taken some reference from geojson) and to fulfil the following objective:

> The value of the properties member is an object (any JSON object or a JSON null value).

For the issue so raised, I think need to modify the function   `  val v : LineString.t array -> t` in `_intf` file and in general the `val v` func of every geometry types. I wonder or if the changes have to be made in `g` func ...
so need your suggestions and guidance on what changes are required to be done in these...

Requesting your review on the latest commit so pushed... :pray: 

And I think then it's almost done and would be great if somehow this PR can be merged today

patricoferris:
> I have made some changes to the properties field as per my understanding. I have changed the properties type as
properties : json option; since I don't find any harm in doing so (taken some reference from geojson) and to fulfil the following objective:
> 
> The value of the properties member is an object (any JSON object or a JSON null value).

I think the interface would be better if it still was a `string * json list` and to take into account the `null` we could make it an `option`. Otherwise people could provide something like `` `String "hello"`` which afaict is not a valid property. See for instance the Rust implementation: https://docs.rs/topojson/latest/topojson/struct.Geometry.html 

```rust
Option<Map<String, JsonValue>>
```
which for us could be:

```ocaml
((string * json) list) option 
```

patricoferris:
> For the issue so raised, I think need to modify the function val v : LineString.t array -> t in _intf file and in general the val v func of every geometry types.

I don't quite follow this, why does it need to change ?

IIITM-Jay:
1. Actually, I have used this before only that you mentioned:
`((string * json) list) option  `

**but was getting the below error, that's why use `json option` instead as we are using  `J.find json [ "properties" ]` to match the `properties` key**  but now :
```
269 |                   { geometry = Point g; properties; foreign_members = fm })
                                              ^^^^^^^^^^
Error: This expression has type json option
       but an expression was expected of type (string * json) list option
       Type json is not compatible with type (string * json) list 
```

IIITM-Jay:
And for this, 


> > For the issue so raised, I think need to modify the function val v : LineString.t array -> t in _intf file and in general the val v func of every geometry types.
> 
> I don't quite follow this, why does it need to change ?

I am getting the error :
```
282 |                     geometry = MultiLineString g;
                                                     ^
Error: This expression has type
         ?properties:json ->
         ?foreign_members:(string * json) list -> geometry -> t
       but an expression was expected of type
         MultiLineString.t = LineString.t array
```

For that reason, I thought so,...
May I request to kindly review the code changes (for the morning and evening commits pushed today) once to help me know what mistakes I was doing

IIITM-Jay:
I have worked on `Polymorphism` concept with other languages but now got crystal clear on how to do it with real world `OCaml`.
Loved the way how `properties` got fixed up and implemented. 

Thanks a lot Patrick... learning more on `OCaml` Thank you so much :pray:

patricoferris:
Thanks @IIITM-Jay for this and for persevering, it's a great piece of work.

For the future, it's a good practice to preserve authorship of code by using tools like `git cherry-pick` and merging code like https://github.com/IIITM-Jay/ocaml-topojson/pull/1. This also helps when trying to track down changes made and who made them at a later date, particularly if the code isn't straightforward.

IIITM-Jay:
Actually, sorry @patricoferris for the above comment, I was literally trying to use ` git cherry-pick` for most of the changes that you've made earlier but everytime I am getting errors and somehow it didn't work. but will take care from now onwards.

But anyways thanks a lot once again :pray:

patricoferris:
> Actually, sorry @patricoferris for the above comment, I was literally trying to use git cherry-pick for most of the changes that you've made earlier but everytime I am getting errors and somehow it didn't work. but will take care from now onwards.

No worries, please ask questions if this is the case though so I can help :))

Issue #5, opened 2022-08-02
Add MDX readme


patricoferris:
It would be good to show off the library in the README using [MDX](https://ocaml.org/p/mdx/2.1.0) à la https://github.com/ocaml-multicore/eio

- [x] Simple read example #6 
- [ ] Reading raw JSON as a string
- [ ] Writing a TopoJSON OCaml value to JSON
- [ ] API examples (e.g. reading some foreign members)

patricoferris:
I've opened separate issues for each of the tasks.

PR #6, opened 2022-08-04
README updates using MDX

IIITM-Jay:
This PR is in continuation with #5.


IIITM-Jay:
Hi @patricoferris , I came up with this PR for the `README.md` updates.

I really found the tool quite interesting, and I have included some of its syntax too. But not found that its necessary, though I used to give it a try.

Looking forward to hear what else we can include and what parts must be removed. (I really felt that the code parts are not needed wherever I mentioned in the file)

IIITM-Jay:
For fixing the issues, I quite surprised as per the syntax included [here](https://github.com/realworldocaml/mdx#file-sync), it says we only have to use `file` LABELS and then the content should automatically comes up and it does so also but it does not automatically corrects the `README.md` file instead it generates a corrected version of it in `_default`.
so, I copy pasted the code part and the issue got removed.

But I think it should automatically correct the original `README.md` file just as how the `dune build @fmt --auto` works( it automatically updates the file)

README.md
```diff
@@ -1,2 +1,94 @@
...
+* [Current Status](#current-status)
+  * [Platform Support](#platform-support)
+  * [Feature Status](#feature-status)
+* [Structure of the Code](#structure-of-the-code)
+
+
+## Introduction
+TopoJSON is a Javascript Object Notation (JSON) format used to represent common topology geographic data structures. TopoJSON, as a GeoJSON extension, supports the following geometry types: "Point", "LineString", "Polygon", "MultiPoint", "MultiLineString", "MultiPolygon", and "GeometryCollection". 
+
+In comparison to geometry,wherein every form is encoded with separate (and regularly redundant) arrays of coordinates, a topology encodes sequences of coordinates in line fragments referred to as arcs that may be shared. For instance, the border between *Massachusetts and New Hampshire*,  is an arc shared by both polygons.
```
patricoferris: I feel like this has been copy and pasted from somewhere? Let's write our own shorter introduction and link out to this if it has come from somewhere else?

README.md
```diff
@@ -1,2 +1,94 @@
...
+
+## Motivation
+Geometries in TopoJSON files are stitched together from shared line segments called arcs, rather than representing them discretely. Arcs are **point sequences**, whereas line strings and polygons are **arc sequences**. Each arc is defined only once, but it can be referenced multiple times by different shapes, ***reducing redundancy and reducing the size of the geospatial data file.*** Hence, the primary advantage of a topology is that it enhances shape simplicity by averting artifacts caused by independently simplifying shapes.
```
patricoferris: If this has also been copied from somewhere, it is better to paraphrase and link out to wherever it came from to maintain correct authorship and attribution.

README.md
```diff
@@ -1,2 +1,94 @@
...
+1. This library is capable of parsing a TopoJSON file as a whole.
+
+2. It supports both the TopoJSON objects - **Topology** and **Geometry**
+
+3. It underpins all the mandatory members of the *Topology* object - **types** and **arcs**.
+
+4. Other than the requisite fields to be supported by all the  *Geometry Objects* i.e., the ***type*** and the ***coordinates/ arcs*** fields, it also supports the following members:
+  - Properties : A geometry object can also additionally have a member with the name “properties”. The value of the properties field is an object (any JSON object or a JSON null value).
+
+<!-- $MDX file=./src/topojson/topojson.ml,part=properties -->
```
patricoferris: I'd rather not use MDX to pull out parts of the implementation code, people can read the files if they're interested. Instead I'd rather actually show examples of using the library with code-blocks. See the `README.md`, the toplevel `dune` file and the `doc/prelude.ml` in this repo https://github.com/patricoferris/hilite as an example of how to do that.

patricoferris (commented):
Great stuff @IIITM-Jay !

MDX will be run with `dune runtest` and you can check that the diff looks good and if it does, you run `dune promote` -- you don't need to manually copy things.

IIITM-Jay:
> Great stuff @IIITM-Jay !

Thank you so much @patricoferris ! but not indeed though  a great content so far... :disappointed: 
 
>  you run `dune promote` -- you don't need to manually copy things.

Got the point as I also read this in the docs too, and I think I had run this command ..._(but don't know if I've probably missed)_

I have followed up with your valuable comments and will be working on the following points:
1. Actually some parts are from _`TopoJSON specs`_  itself and some of them are from some sites when I visited them to learn a bit more  and found some great stuffs. But before pushing this PR I should honestly say that I paraphrased it by my own and also used some tools also but didn't get any satisfactory output and sounded more or less similar to original one.
_**So it would be best here though, that we can have our own content and documentation for this library.
Hence, I will come up with some good stuff and  will also incorporate the valuable suggestions**_ 

> Let's write our own shorter introduction

so yep, indeed will work on this .... :+1:

IIITM-Jay:
2. Secondly, for this, to show _**examples and use cases**_ sort of stuffs
> Instead I'd rather actually show examples of using the library with code-blocks.

as I mentioned earlier that I have used MDX to just give it a try and wanted to know how efficiently we can use this to document this library. 
**_So I think it would be better if we can include like how this library would be used to read a `TopoJSON file` or more stuff like that, a sort of `parsing a file` into OCaml representation what else we can include? need some more valuable suggestions & guidance on the same..._**

IIITM-Jay:
Hi @patricoferris, updated the contents of `Introduction` and `Motivation` sections. 
Wanted to know more on how can we efficiently write the `Examples` section on how this lib can be used to read/ write/ parse a `json` file..., not getting exactly. I am confused as are we in a position to write something like in the _example section_:

```ocaml
# # require "topojson" ;;
``` 
An example here can be helpful to write this section...

README.md
```diff
@@ -1,2 +1,67 @@
-# ocaml-topojson
-A pure OCaml library for working with the TopoJSON format
+# ocaml-topojson --- OCaml library to work with TopoJSON Format Specification
+
+__Abstract__
```
patricoferris:
```suggestion
```

README.md
```diff
@@ -1,2 +1,67 @@
...
+* [Motivation](#motivation)
+* [Current Status](#current-status)
+  * [Platform Support](#platform-support)
+  * [Feature Status](#feature-status)
+* [Structure of the Code](#structure-of-the-code)
+* [Examples](#examples)
+
+
+## Introduction
+TopoJSON is an enhanced format for encoding GeoJSON geospatial data. In addition to the GeoJSON geometry types, viz. "Point", "LineString", "Polygon", "MultiPoint", "MultiLineString", "MultiPolygon", and "GeometryCollection", TopoJSON instigates a very new type "Topology", which comprises of GeoJSON objects. A _topology_ has a field `objects` mapped with one or more geometry objects by its name type.
```
patricoferris:
```suggestion
TopoJSON is an enhanced format for encoding GeoJSON geospatial data. In addition to the GeoJSON geometry types, viz. "Point", "LineString", "Polygon", "MultiPoint", "MultiLineString", "MultiPolygon", and "GeometryCollection", TopoJSON instigates a new type "Topology", which comprises of GeoJSON objects. A _topology_ has a field `objects` mapped with one or more geometry objects by its name type. 
```

README.md
```diff
@@ -1,2 +1,67 @@
...
+
+### TopoJSON Vs GeoJSON
+TopoJSON has an `arcs` member which consists of the `coordinates` of the geometry types (except "Point" and "MultiPoint"). The value of the _"arcs"_ member is an array of arrays of positions. This is a primary advantage over GeoJSON objects where each individual geometries have their own separately defined coordinates.
+See [TopoJSON Format Specification](https://github.com/topojson/topojson-specification) to know more.
+
+
+## Motivation
+Currently OCaml's [GeoJSON Library](https://github.com/geocaml/ocaml-geojson) supports various types of geometry objects as being laid down by the [IETF RFC 7946 - The GeoJSON Format Specification](https://www.rfc-editor.org/rfc/rfc7946.html). A similar sort of library is being implemented to  support *topojson specification*.
```
patricoferris: Let's remove this paragraph

README.md
```diff
@@ -1,2 +1,67 @@
...
+
+### TopoJSON Vs GeoJSON
+TopoJSON has an `arcs` member which consists of the `coordinates` of the geometry types (except "Point" and "MultiPoint"). The value of the _"arcs"_ member is an array of arrays of positions. This is a primary advantage over GeoJSON objects where each individual geometries have their own separately defined coordinates.
+See [TopoJSON Format Specification](https://github.com/topojson/topojson-specification) to know more.
+
+
+## Motivation
+Currently OCaml's [GeoJSON Library](https://github.com/geocaml/ocaml-geojson) supports various types of geometry objects as being laid down by the [IETF RFC 7946 - The GeoJSON Format Specification](https://www.rfc-editor.org/rfc/rfc7946.html). A similar sort of library is being implemented to  support *topojson specification*. 
+
+TopoJSON helps to reduce the size of the geospatial data file in a way by eradicating the redundancy and eliminating the duplicate topology that is being shared by one or more geometries. For example, a common boundary that is being shared between two states/ countries can be represented only once and can be referenced multiple times.
```
patricoferris:
```suggestion
TopoJSON helps to reduce the size of the geospatial data file in a way by eradicating the redundancy and eliminating the duplicate topology that is being shared by one or more geometries. For example, a common boundary that is being shared between two states/countries can be represented only once and can be referenced multiple times.
```

README.md
```diff
@@ -1,2 +1,67 @@
...
+To know more about [TopoJSON](https://github.com/topojson/topojson) and its related advantages.
+
+## Current status
+### Platform Support
```
patricoferris: Let's remove the platform status section

dune
```diff
@@ -0,0 +1,3 @@
+(mdx
+ (packages topojson)
```
patricoferris, line 2:
```suggestion
 (packages topojson)
 (preludes doc/prelude.ml)
```

doc/prelude.ml
```diff
@@ -0,0 +1 @@
+#require "topojson"
```
patricoferris:
```suggestion
#require "topojson";;
```

patricoferris:
I've completed the setup in this commit -- https://github.com/geocaml/ocaml-topojson/commit/575fdfb474c28386626dd2e695aa95bac7aa03bd

In particular I renamed the prelude file so ocamlformat doesn't touch it. It also adds the Ezjsonm parser.

IIITM-Jay:
Hi @patricoferris, after spending a lot of time on resolving conflicts and errors with `cherry-pick`, I've finally did it.
but wondering if have I done it correctly?

patricoferris:
Seems to be fine, when using `cherry-pick` it is like rebasing so if there are conflicts, you should be able to fix them as you go. I'm not quite sure why there were conflicts though, because https://github.com/geocaml/ocaml-topojson/commit/575fdfb474c28386626dd2e695aa95bac7aa03bd is built on top of https://github.com/geocaml/ocaml-topojson/commit/07a006299f25e10be01fe00eea090b15a52fa0b0 so it should have just worked (perhaps you didn't pull those changes in before trying to cherry-pick the commit). Anyway, you fixed the conflicts in subsequent commits so that's fine, now we just need the MDX examples :))

IIITM-Jay:
I wonder if we change :
```ocaml
# Topojson.of_json (`String "I should fail!");;
```
something like
```ocaml
# Topojson.of_json (`String "Topology");;
```
it throws the same error...
 ```ocaml
 # Topojson.of_json (`String "Topology");;
-
+- : (Topojson.t, [ `Msg of string ]) result =
+Error (`Msg "Could not find Topology type")
 ```

IIITM-Jay:
Am I misunderstood the above comment in anyways...

shall we have to use `Topojosn.to_json` in the `Writing` section?

patricoferris:
```ocaml
# Topojson.of_json (`String "Topology");;
-
+- : (Topojson.t, [ `Msg of string ]) result =
+Error (`Msg "Could not find Topology type")
```

This is completely expected. A TopoJSON object is a JSON object not a string so it will always throw this error if you try and parse a JSON string (not to be confused with a JSON object encoded as a string).

It would be good to add a "reading" example that also works and a "writing" example using `to_json`. You might need to add more constructors to the interface to allow you to build a TopoJSON object.

IIITM-Jay:
1. I have added one more `error` example that I thought it should be mentioned.
**_For the correct example I tried something like:_**
```ocaml
("type", `String "Topology"); 
("objects" , `O [ ]); 
("arcs", `Array
      [| [| [| 102.; 0. |]; [| 103.; 1. |]; [| 104.; 0. |]; [| 105.; 1. |] |] |] 
)     
```
but got the error, how can we resolve this and also need some sort of suggestions  on how to write good example though:
```ocaml
+Error: This expression has type [> `Array of float array array array ]
+       but an expression was expected of type Topojson.json
+       The second variant type does not allow tag(s) `Array
``` 


2. > You might need to add more constructors to the interface to allow you to build a TopoJSON object.

Sorry but wanted to get a little bit of elaboration on the above statement, need some more guidance on the same regarding the **_`writing` section_** .

patricoferris:
I don't think we need erroneous examples, I just added that to show how it should be done in terms of using MDX to run OCaml code rather than copy code out of files.

Your example is failing because the variant is called `` `A`` and not `` `Array`` https://github.com/mirage/ezjsonm/blob/master/lib/ezjsonm.ml#L23 for describing JSON arrays with Ezjsonm.

For writing you need to construct a TopoJSON OCaml value (using functions like `<Module>.v`) and then convert it to JSON.

IIITM-Jay:
Actually, for the very first time I used `` `A `` but was throwing the error _(as mentioned below)_, then I thought may be I am wrong so tried with `` `Array `` _( with both lowercase and uppercase letter A)_ , but that also didn't work and again that you mentioned about the variant I tried once again so here is what I am getting:

```ocaml
# Topojson.of_json ( `O [ 
  ("type", `String "Topology"); 
  ("objects" , `O [ ("type" , `String "Polygon"); ("arcs", `A  [| [| 1 |] |])  ]); 
  ("arcs", `A  
  [|
      [| [| 102.; 0. |]; [| 103.; 1. |]; [| 104.; 0. |]; [| 105.; 1. |] |];
      [|
        [| 100.; 0. |];
        [| 101.; 0. |];
        [| 101.; 1. |];
        [| 100.; 1. |];
        [| 100.; 0. |];
      |];
  |] )
  ]);;    
Line 3, characters 64-77:
Error: This expression has type 'a array
       but an expression was expected of type Topojson.json list
```
and as I can see the `` `A `` is of type `value list` so for that also I got a bit confused as to whether is this compatible with `array` type _(as converting to `list` also it says  **``` expression was expected of type Topojson.json```**)_

patricoferris:
The error message let's us know all the problems:

> Error: This expression has type 'a array
       but an expression was expected of type Topojson.json list

1. You have provided an _OCaml_ array to the constructor `` `A`` when the constructor expects an OCaml list i.e. `[ 1. ]` not `[| 1. |]`. 
2. The list given to `` `A`` should contain _JSON OCaml_ values not plain OCaml values. You have used `1.`, an OCaml `float` but it should be a JSON value i.e. `` `Float 1.``

patricoferris:
Thanks @IIITM-Jay -- I took the liberty of merging now, feel free to open a new PR if you wish to add a writing example :))

PR #10, opened 2022-10-12
Added ID fields to geometry object

### LINKED ISSUE #4: Geometry objects can contain `"id"` fields
The TopoJSON spec mentions an `"id"` field, similarly to GeoJSON which should be handled by this library:

> If a geometry has a commonly used identifier, that identifier should be included as a member of the geometry object with the name “id”. 

(see https://github.com/topojson/topojson-specification#22-geometry-objects)

mimi-tech:
Fixes #4

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,4 +1,4 @@
-version = 0.24.1
+version = 0.21.0
 profile = conventional
 break-infix = fit-or-vertical
 parse-docstrings = true
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,4 +1,4 @@
-version = 0.21.0
+version = 0.24.1
 profile = conventional
 break-infix = fit-or-vertical
 parse-docstrings = true
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -248,15 +248,16 @@ module Make (J : Intf.Json) = struct
       geometry : geometry;
       properties : properties;
       foreign_members : (string * json) list;
-      id : id;
+      id : json option;
     }
 
     let v ?(properties = `None) ?(foreign_members = []) geo =
-      { geometry = geo; properties; foreign_members }
+      { geometry = geo; properties; foreign_members; id = None }
 
     let geometry t = t.geometry
     let properties t = t.properties
     let foreign_members t = t.foreign_members
+    let id t = t.id
 
     let properties_of_json json =
       match J.find json [ "properties" ] with
@@ -271,15 +272,30 @@ module Make (J : Intf.Json) = struct
           match J.to_string typ with
           | Ok "Point" ->
               Result.map (fun g ->
-                  { geometry = Point g; properties; foreign_members = fm })
+                  {
+                    geometry = Point g;
+                    properties;
+                    foreign_members = fm;
+                    id = None;
+                  })
               @@ Point.base_of_json json
           | Ok "MultiPoint" ->
               Result.map (fun g ->
-                  { geometry = MultiPoint g; properties; foreign_members = fm })
+                  {
+                    geometry = MultiPoint g;
+                    properties;
+                    foreign_members = fm;
+                    id = None;
+                  })
               @@ MultiPoint.base_of_json json
           | Ok "LineString" ->
               Result.map (fun g ->
-                  { geometry = LineString g; properties; foreign_members = fm })
+                  {
+                    geometry = LineString g;
+                    properties;
+                    foreign_members = fm;
+                    id = None;
+                  })
               @@ LineString.base_of_json json
           | Ok "MultiLineString" ->
               Result.map (fun g ->
@@ -287,11 +303,17 @@ module Make (J : Intf.Json) = struct
                     geometry = MultiLineString g;
                     properties;
                     foreign_members = fm;
+                    id = None;
                   })
               @@ MultiLineString.base_of_json json
           | Ok "Polygon" ->
               Result.map (fun g ->
-                  { geometry = Polygon g; properties; foreign_members = fm })
+                  {
+                    geometry = Polygon g;
+                    properties;
+                    foreign_members = fm;
+                    id = None;
+                  })
               @@ Polygon.base_of_json json
           | Ok "MultiPolygon" ->
               Result.map (fun g ->
@@ -299,6 +321,7 @@ module Make (J : Intf.Json) = struct
                     geometry = MultiPolygon g;
                     properties;
                     foreign_members = fm;
+                    id = None;
                   })
               @@ MultiPolygon.base_of_json json
           | Ok "GeometryCollection" -> (
@@ -311,6 +334,7 @@ module Make (J : Intf.Json) = struct
                         geometry = Collection g;
                         properties;
                         foreign_members = fm;
+                        id = None;
                       })
                     geo
               | None ->
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -26,8 +26,6 @@ module Make (J : Intf.Json) = struct
       | `Null -> [ ("properties", J.null) ]
       | `Obj v -> [ ("properties", J.obj v) ]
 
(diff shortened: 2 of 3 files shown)


mimi-tech:
@patricoferris Please can you review and let me know what you think. Thank you.

.ocamlformat
```diff
@@ -1,4 +1,4 @@
-version = 0.21.0
+version = 0.24.1
```
patricoferris:
```suggestion
version = 0.21.0
```

It probably makes sense to upgrade at some point, but not in this PR.

src/topojson/topojson.ml
```diff
@@ -248,6 +248,7 @@ module Make (J : Intf.Json) = struct
       geometry : geometry;
       properties : properties;
       foreign_members : (string * json) list;
+      id : id;
```
patricoferris: Adding an `id` field to this record seems like a good choice but there are some issues.

1. You have given it the type `id` but no such type exists. It probably should be of type `json` and it should also be optional. The spec doesn't _require_ and `id` field.
2. Anywhere we subsequently construct a geometry record needs to be update (e.g. in the function `v` directly afterwards)
3. The JSON encoders (e.g. `Geometry. base_of_json`) will need updating too.

patricoferris (changes requested):
Thanks @mimi-tech! A few general comments for PRs in the future. It is good practice to provide a description of what the PR does, what you think it fixes (I updated the description to that) and any questions you might have about the code you are contributing. For example, as the CI indicates, this PR doesn't compile. Why do you think that is? (hint: I've mentioned it in my comments)?

In general it's a good idea to submit code that compiles locally unless you are really stuck. If you are stuck then submitting code that doesn't compile is completely okay, but you should then be asking questions and explaining why you think it isn't working (e.g. in this case you could say "I wasn't sure what type to give the `id` field" -- for me, this makes it much clearer that you understand what `x : y` means in a record, but you just weren't sure what `y` to use)

mimi-tech:
@patricoferris Thanks for the reply. Sorry for not asking properly. I thought you meant I should go ahead and do it as I stated for you before making changes to the code which is these 👇 

and t = {
geometry : geometry;
properties : properties;
foreign_members : (string * json) list;
id:id;

src/topojson/topojson.ml
mimi-tech: @patricoferris Thank you for clarifying me
1. Line 251 will be replaced with id : json;
2. line 254 to 259  will become
```javascript
let v ?(properties = `None) ?(foreign_members = []) ?(id={}) geo =
      { geometry = geo; properties; foreign_members }

    let geometry t = t.geometry
    let properties t = t.properties
    let foreign_members t = t.foreign_members
    let id t = t.id

```

Please correct me and please which command will I use to compile locally before making a PR. Thank you

mimi-tech:
@patricoferris Please what do I pass here I ran the test command. It gave me an error "Syntax error" pointing at id = {}
let v ?(properties = `None) ?(foreign_members = []) ?(id = {}) geo =
      { geometry = geo; properties; foreign_members; id }

Please what do I pass there. If I am correct else please direct me. Thank you.

src/topojson/topojson.ml
patricoferris: Great :)) Note that in general I'm not telling you _everything_ you should or shouldn't do as it's good for you to discover these things yourself as well, and in general I'll try to link to documentation rather than tell you directly to help you discover answers yourself. Please don't hesitate to ask questions though!

Re (1) -- It should probably be optional, OCaml has optional types. See https://cs3110.github.io/textbook/chapters/data/options.html. This should also help with (2). From your code (you can format code with [codeblocks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks) in Github) it looks like the record is still missing it's `id` value.

See the contributing guide in the root of this repo for how to build the code.

patricoferris:
`{}` is not valid OCaml syntax, there's no such thing as an "empty record" in OCaml. Instead of an empty record we can have optional values of type `'a option` (pronounced "alpha option") which is either `Some v` where `v` is of type `'a` or `None` to represent nothing. When used with functions as "optional arguments" is looks like https://dev.realworldocaml.org/variables-and-functions.html#scrollNav-2-7

src/topojson/topojson.ml
mimi-tech: @patricoferris  I ran the test command and it gave me an error   { geometry = Point g; properties; foreign_members = fm })
Error: Some record fields are undefined: id
Please how do I define the id. Thank you.

Line 275 will be 
```javascript
 Some typ -> (
          match J.to_string typ with
          | Ok "Point" ->
              Result.map (fun g ->
                  { geometry = Point g; properties; foreign_members = fm; id = })
              @@ Point.base_of_json json

```

The base_of_json json  function

mimi-tech: @patricoferris Please can you reply 👆 Thank you

mimi-tech:
Hi @patricoferris. Please these is what I have so far

```javascript

and t = {
      geometry : geometry;
      properties : properties;
      foreign_members : (string * json) list;
      id : json option;
    }

    let v ?(properties = `None) ?(foreign_members = []) ?(id = `a) geo =
      { geometry = geo; properties; foreign_members; id }

```

I am getting this error
File "src/topojson/topojson.ml", line 255, characters 53-55:
255 |       { geometry = geo; properties; foreign_members; id }
                                                                                                  ^^
Error: This expression has type [> `a ]
       but an expression was expected of type json option

Please help me on what to do. Thank you.

patricoferris:
From the section in RWO that I linked:

> Here, ? is used in the definition of the function to mark sep as optional. And while the caller can pass a value of type string for sep, internally to the function, sep is seen as a string option, with None appearing when sep is not provided by the caller.

So it should be enough to add `?id` which means if the argument is not supplied it will be given the value `None`

mimi-tech:
@patricoferris. Thanks for replying. Please will this file be updated too "topojson_intf.ml" in fixing this bug. Thank you.

patricoferris:
Yep, the [interfaces](https://ocaml.org/docs/modules#interfaces-and-signatures) will need updating 👍

patricoferris (commented):
This is looking great! All that's left is to actually get the `id` value. I think you could copy what we do for `properties` for the most part

src/topojson/topojson.ml
```diff
@@ -26,6 +26,8 @@ module Make (J : Intf.Json) = struct
       | `Null -> [ ("properties", J.null) ]
       | `Obj v -> [ ("properties", J.obj v) ]
 
+    type id = [ `None | `Null | `Obj of (string * json) list ]
```
patricoferris: Sorry, I meant the properties should be taken as _inspiration_ not copied ^^" -- I don't think we need to create a new polymorphic variant for this. Using `json option` should be sufficient.

src/topojson/topojson.ml
```diff
@@ -248,56 +250,76 @@ module Make (J : Intf.Json) = struct
...
     let properties t = t.properties
     let foreign_members t = t.foreign_members
+    let id t = t.id
 
     let properties_of_json json =
       match J.find json [ "properties" ] with
       | Some j -> if J.is_null j then `Null else `Obj (decode_or_err J.to_obj j)
       | None -> `None
 
+    let id_of_json json =
```
patricoferris: Same here -- the topojson specification doesn't make it clear what possible values the "id" should take, so we shouldn't narrow it down to just an object or null. `J.find json [ "id" ]` should suffice

patricoferris (changes requested):

patricoferris (approved):
Excellent, I pushed a commit to re-expose the `id` argument and tidy up the docs, but if the CI is green this is good to merge. Thank you :))

mimi-tech:
Hi @patricoferris, Hope you enjoyed your weekend. I am so grateful for you assistance and making me to understand OCaml. You are the best 😍🤗. Please can you assign me another task? Thank you.

PR #11, opened 2022-10-12
Fix foreign members

Priceless-P:
I fixed the `foreign members` implementation. I made it so that `Point` and `Multipoint` does not include `arcs`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,29 +2,32 @@
 
 ## OCaml and opam
 
-After following [these instructions for setting up OCaml with opam](https://v3.ocaml.org/docs/up-and-running) you should be in a position to start building this
-repository. Refer to the following section for cloning the repository using git.
+After following [these instructions for setting up OCaml with opam](https://v3.ocaml.org/docs/up-and-running) you should be in a position to start building this repository. Refer to the following section for cloning the repository using git.
 
-Once you have that set up to install the dependencies you can run 
+Once you have done the set up, next install the dependencies. 
+
+To install the dependencies, you run: 
 
 ```
 opam install . --deps-only --with-test
 ```
 
-This will install the main dependencies along with any we will need for running the tests. The build the project after opam has finished installing everything run:
+This will install the main dependencies along with any other, which will be needed for running the tests. 
+
+To build the project, once opam has finished installing everything, run:
 
 ```
 dune build
 ```
 
-And to run the tests (which are in the `test` directory) run:
+Also,to run the tests (which are in the `test` directory) run:
 
 ```
 dune runtest
 ```
 
 
-After making changes to the code please also run the formatter to maintain a common style across the codebase. You can achieve this by running the following command:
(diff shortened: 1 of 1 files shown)


src/topojson/topojson.ml
```diff
@@ -262,19 +262,26 @@ module Make (J : Intf.Json) = struct
       | Some j -> if J.is_null j then `Null else `Obj (decode_or_err J.to_obj j)
       | None -> `None
 
+    let new_key_in_use = [ "arcs" ]
+
     let rec base_of_json json =
       let fm = foreign_members_of_json json in
+      let foreign = List.filter (fun (k, _v) -> List.mem k new_key_in_use) fm in
```
patricoferris: This is starting to head in the right direction but isn't quite right. At the moment the code does the following:

1. Calculates the foreign members by keeping everything that doesn't match `keys_in_use` -- this means `coordinates` and `arcs` are now longer in the value `fm`.
2. Your code then filters `fm` to only keep values that have a key in `new_key_in_use` i.e. `"arcs"` but given (1) this will never be the case and so `foreign` will _always_ be the empty list.

Consider this simplified code of what's happening:

```ocaml
let keys_in_use = [ "arcs" ]
let new_key_in_use = [ "arcs" ]

List.filter (fun (k, _) -> not (List.mem k keys_in_use)) [ ("arcs", 1) ]
|> List.filter (fun (k, _v) -> List.mem k new_key_in_use)
```
Which returns `- : (string * int) list = []`.

I think this might be easier to make the `X.base_of_json` return a pair `(a, b)` where `a` is the current return value (the geometry) and the second is the foreign members.

patricoferris (changes requested):
Thanks @Priceless-P -- the logic isn't quite right for fixing the issue.

src/topojson/topojson.ml
Priceless-P: @patricoferris If I try to change the return type of `Point` and `MultiPoint`  to `(a, b)` OCaml complains that return type of the rest are not compatible with the expected type 🥲  

```
This expression has type (t, [ `Msg of string ]) result
but an expression was expected of type
  ((t -> geometry) * (json -> (string * json) list), [ `Msg of string ])
  result
Type t is not compatible with type
  (t -> geometry) * (json -> (string * json) list)
```

patricoferris: Hmm, I meant changing `Point.base_of_json` to return a pair so you might have:

```ocaml
Result.map (fun (g, fm) ->
      { geometry = Point g; properties; foreign_members = fm })
  @@ Point.base_of_json json
```

Alternatively, you could make `foreign_members_of_json` more flexible.

```ocaml
let fm = foreign_members_of_json true json in
  Result.map (fun g ->
      { geometry = Point g; properties; foreign_members = fm })
  @@ Point.base_of_json json
```

patricoferris: Note you might need to rebase on https://github.com/geocaml/ocaml-topojson/pull/12

PR #15, opened 2022-10-20
Fix foreign member

Priceless-P:
I created a new `keys_in_use_for_point` which doesn't contain `arcs` and applied the resultant `foreign_members_of_json_for_point` to `Point` and `MultiPoint` and then pushed the changes to `foreign_member` branch, hence the reason I'm opening a new PR on the same issue

### DIFF
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -54,6 +54,7 @@ module Make (J : Intf.Json) = struct
       | Ok assoc ->
           List.filter (fun (k, _v) -> not (List.mem k keys_in_use)) assoc
       | Error _ -> []
+
     let parse_with_coords json p_c typ =
       match (J.find json [ "type" ], J.find json [ "coordinates" ]) with
       | None, _ ->
@@ -124,7 +125,6 @@ module Make (J : Intf.Json) = struct
       let position = Fun.id
       let v position = position
       let parse_coords coords = J.to_array (decode_or_err J.to_float) coords
-
       let base_of_json json = parse_with_coords json parse_coords typ
 
       let to_json ?bbox ?(properties = `None) ?(foreign_members = []) position =
@@ -326,8 +326,8 @@ module Make (J : Intf.Json) = struct
                     geometry = Polygon g;
                     properties;
                     foreign_members = fm keys_in_use;
-                    id
-                   })
+                    id;
+                  })
               @@ Polygon.base_of_json json
           | Ok "MultiPolygon" ->
               Result.map (fun g ->
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -321,7 +321,12 @@ module Make (J : Intf.Json) = struct
               @@ MultiLineString.base_of_json json
           | Ok "Polygon" ->
               Result.map (fun g ->
-                  { geometry = Polygon g; properties; foreign_members = fm keys_in_use; id })
+                  {
+                    geometry = Polygon g;
+                    properties;
+                    foreign_members = fm keys_in_use;
+                    id
+                   })
               @@ Polygon.base_of_json json
           | Ok "MultiPolygon" ->
               Result.map (fun g ->
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -38,10 +38,29 @@ module Make (J : Intf.Json) = struct
         "geometries";
       ]
 
+    let keys_in_use_for_point =
+      [
+        "type";
+        "properties";
+        "coordinates";
+        "bbox";
+        "id";
+        "objects";
+        "geometries";
+      ]
+
     let foreign_members_of_json json =
       match J.to_obj json with
       | Ok assoc ->
           List.filter (fun (k, _v) -> not (List.mem k keys_in_use)) assoc
+      | Error _ -> []
+
+    let foreign_members_of_json_for_point json =
+      match J.to_obj json with
+      | Ok assoc ->
+          List.filter
+            (fun (k, _v) -> not (List.mem k keys_in_use_for_point))
+            assoc
       | Error _ -> []
 
     let parse_with_coords json p_c typ =
@@ -275,14 +294,25 @@ module Make (J : Intf.Json) = struct
           match J.to_string typ with
           | Ok "Point" ->
               Result.map (fun g ->
-                  { geometry = Point g; properties; foreign_members = fm; id })
+                  let point_foreign_members =
+                    foreign_members_of_json_for_point json
+                  in
+                  {
+                    geometry = Point g;
+                    properties;
+                    foreign_members = point_foreign_members;
+                    id;
+                  })
               @@ Point.base_of_json json
           | Ok "MultiPoint" ->
               Result.map (fun g ->
+                  let point_foreign_members =
+                    foreign_members_of_json_for_point json
+                  in
                   {
                     geometry = MultiPoint g;
                     properties;
-                    foreign_members = fm;
+                    foreign_members = point_foreign_members;
                     id;
                   })
               @@ MultiPoint.base_of_json json
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -49,20 +49,11 @@ module Make (J : Intf.Json) = struct
         "geometries";
       ]
 
-    let foreign_members_of_json json =
+    let foreign_members_of_json json keys_in_use =
       match J.to_obj json with
       | Ok assoc ->
           List.filter (fun (k, _v) -> not (List.mem k keys_in_use)) assoc
       | Error _ -> []
-
-    let foreign_members_of_json_for_point json =
-      match J.to_obj json with
-      | Ok assoc ->
-          List.filter
-            (fun (k, _v) -> not (List.mem k keys_in_use_for_point))
-            assoc
-      | Error _ -> []
-
     let parse_with_coords json p_c typ =
       match (J.find json [ "type" ], J.find json [ "coordinates" ]) with
       | None, _ ->
@@ -294,25 +285,19 @@ module Make (J : Intf.Json) = struct
           match J.to_string typ with
           | Ok "Point" ->
               Result.map (fun g ->
-                  let point_foreign_members =
-                    foreign_members_of_json_for_point json
-                  in
                   {
                     geometry = Point g;
                     properties;
-                    foreign_members = point_foreign_members;
+                    foreign_members = fm keys_in_use_for_point;
                     id;
                   })
               @@ Point.base_of_json json
           | Ok "MultiPoint" ->
               Result.map (fun g ->
-                  let point_foreign_members =
-                    foreign_members_of_json_for_point json
-                  in
                   {
                     geometry = MultiPoint g;
                     properties;
-                    foreign_members = point_foreign_members;
+                    foreign_members = fm keys_in_use_for_point;
                     id;
                   })
               @@ MultiPoint.base_of_json json
@@ -321,7 +306,7 @@ module Make (J : Intf.Json) = struct
                   {
                     geometry = LineString g;
                     properties;
-                    foreign_members = fm;
+                    foreign_members = fm keys_in_use;
                     id;
                   })
               @@ LineString.base_of_json json
@@ -330,20 +315,20 @@ module Make (J : Intf.Json) = struct
                   {
                     geometry = MultiLineString g;
                     properties;
-                    foreign_members = fm;
+                    foreign_members = fm keys_in_use;
                     id;
                   })
               @@ MultiLineString.base_of_json json
           | Ok "Polygon" ->
               Result.map (fun g ->
-                  { geometry = Polygon g; properties; foreign_members = fm; id })
+                  { geometry = Polygon g; properties; foreign_members = fm keys_in_use; id })
               @@ Polygon.base_of_json json
           | Ok "MultiPolygon" ->
               Result.map (fun g ->
                   {
                     geometry = MultiPolygon g;
                     properties;
-                    foreign_members = fm;
+                    foreign_members = fm keys_in_use;
                     id;
                   })
               @@ MultiPolygon.base_of_json json
@@ -356,7 +341,7 @@ module Make (J : Intf.Json) = struct
                       {
                         geometry = Collection g;
                         properties;
-                        foreign_members = fm;
+                        foreign_members = fm keys_in_use;
                         id;
                       })
                     geo
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -124,6 +124,7 @@ module Make (J : Intf.Json) = struct
       let position = Fun.id
       let v position = position
       let parse_coords coords = J.to_array (decode_or_err J.to_float) coords
+
       let base_of_json json = parse_with_coords json parse_coords typ
 
       let to_json ?bbox ?(properties = `None) ?(foreign_members = []) position =
--- a/test/topojson/test.ml
+++ b/test/topojson/test.ml
@@ -61,6 +61,12 @@ let pp_position ppf t =
   let lng = Position.lng t in
   Fmt.pf ppf "[%f, %f]" lat lng
 
+let foreign_member ppf v =
+  let open Topojson.Geometry.Point in
+  let foreign = "arcs" in
+  Fmt.pf ppf "[%s]" foreign
+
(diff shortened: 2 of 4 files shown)


src/topojson/topojson.ml
```diff
@@ -38,12 +38,31 @@ module Make (J : Intf.Json) = struct
...
+        "geometries";
+      ]
+
     let foreign_members_of_json json =
       match J.to_obj json with
       | Ok assoc ->
           List.filter (fun (k, _v) -> not (List.mem k keys_in_use)) assoc
       | Error _ -> []
 
+    let foreign_members_of_json_for_point json =
```
patricoferris: Why not keep one implementation of `foreign_members_of_json` and pass in an extra argument called `keys_in_use` instead?

patricoferris (changes requested):
Thanks! This looks good, just some suggestions to reuse the code.

src/topojson/topojson.ml
Priceless-P: That is true, please take a look at the changes i made. Is it correct?

patricoferris:
Looks good to me! Thank you :)) Two last things:

 - To format the code, there's an explanation in the CONTRIBUTING document. But you can run `dune build @fmt --auto` after installing the _correct_ version of ocamlformat.
 - It would be _very_ good to add a test of the foreign member logic differences to make sure it is working properly and doesn't break in the future.

Priceless-P:
>To format the code, there's an explanation in the CONTRIBUTING document. But you can run dune build @fmt --auto after installing the correct version of ocamlformat.

Yes, It skipped my mind as a result of excitement that I might I have finally figured it out 😂. I'll make sure to remember it always from now on. I have also added the test of the foreign member. I ran `dune runtest` and it was successful. I believe that is the desired result.

Thank you so much for your help @patricoferris

patricoferris:
Nearly there :)) Can you just check in the actual test parsing code that `arcs` is treated as a foreign member for the point?

Priceless-P:
Hi @patricoferris , this is what i have in test/topojson/test_cases/files/examplar.json
```
{
          "type": "Point",
          "properties": {
            "prop0": "value0"
          },
          "coordinates": [102, 0.5],
          "arcs": [0]
        },
```
I added `"arcs: [0]"` to the point object. But when i run `dune runtest`, it passes the first time but doesn't seem to respond if i try to run it again. Please what could the issue be?

patricoferris:
Dune is clever, if no files have changed it won't rerun anything because it won't change anything. I saw the changes you made, but the old code would also still be able to read the file. You will need to make changes in the test code itself https://github.com/geocaml/ocaml-topojson/blob/main/test/topojson/test.ml

Priceless-P:
>Dune is clever, if no files have changed it won't rerun anything because it won't change anything. 

Wow. That's cool!

>You will need to make changes in the test code itself

Okay. So I'm thinking that the steps I need to take to test this are:
- Find a way to get to the `Point` Object in examplar.json
- Ensure that if `"arcs"` is passed it is treated as a foreign member maybe return `NULL` so it is not taken into account.

@patricoferris Please is this correct? Can you give me some tips on the best way achieve this?

Priceless-P:
After looking at the  existing code in `test.ml`. I think i need to do something like
```
let foreign_member ppf v =
let open Topojson.Geometry.Point in
let foreign = "arcs" in
Fmt.pf ppf "[%s]" foreign
let foreign_members = Alcotest.testable foreign_member Stdlib.( = )

let expected_foreign_members = "arcs"
```

Then in the `main()` function I will add something like
```
Alcotest.(check (list (list foreign_members))) "same foreign_member" f.foreign_members expected_foreign_members;
```
My solution is not working. I know I'm not getting the value right plus I have several type incompatibility errors. Please is this the right way to approach this task? If so, how do I get rid of the errors?

patricoferris:
This, from a high-level, looks like the right approach! Please do push your code even if it isn't compiling because it's quite hard to know exactly what code you have added and what compiler problems you are facing :))

Priceless-P:
@patricoferris, Thank you for replying. I have pushed the code.

test/topojson/test.ml
```diff
@@ -61,6 +61,12 @@ let pp_position ppf t =
   let lng = Position.lng t in
   Fmt.pf ppf "[%f, %f]" lat lng
 
+let foreign_member ppf v =
```
patricoferris: Here you are trying to create a pretty printer for foreign members, this means the signature should look like:

```
val pp_foreign_member : Format.formatter -> (string * Ezjsonm.value) list -> unit
```

So you will probably need something along the lines of 

```ocaml
let pp_foreign_member ppf (v : (string * Ezjsonm.value) list) =
  Fmt.pf ppf "%a" Fmt.(list (pair string pp_ezjsonm)) v
```

test/topojson/test.ml
```diff
@@ -92,6 +98,8 @@ and remove acc k = function
   | (k', _) :: rest when k = k' -> List.rev acc @ rest
   | x :: rest -> remove (x :: acc) k rest
 
+let expected_foreign_members = "arcs"
```
patricoferris: Foreign members are expected to be of type `(string * Ezjsonm.value) list`.

test/topojson/test.ml
```diff
@@ -106,6 +114,8 @@ let main () =
       (* Here we check that the arcs defined in the file are the same as the ones
          we hardcoded above *)
       Alcotest.(check (array (array position))) "same arcs" f.arcs expected_arcs;
+      Alcotest.(check (list (list foreign_members)))
+        "same foreign_member" f.foreign_members expected_foreign_members;
```
patricoferris: Expected comes before the value you are testing

patricoferris (changes requested):
Note when you are getting the foreign members, you are getting them for the top-level topology object but you wanted I think the topology -> geometry collection -> point object. I think it might be worthwhile adding a new simpler topojson document to the test_case files and using that to test this PR

test/topojson/test.ml
Priceless-P: @patricoferris, I have corrected this. Did I get it right?

test/topojson/test.ml
```diff
@@ -61,12 +61,10 @@ let pp_position ppf t =
   let lng = Position.lng t in
   Fmt.pf ppf "[%f, %f]" lat lng
 
-let foreign_member ppf v =
-  let open Topojson.Geometry.Point in
-  let foreign = "arcs" in
-  Fmt.pf ppf "[%s]" foreign
+  let pp_foreign_member ppf (v : (string * Ezjsonm.value) list) =
+    Fmt.pf ppf "%a" Fmt.(list (pair string pp_ezjsonm)) v
```
patricoferris: You will need to provide an implementation for `pp_ezjsonm` I left it out of my review :))

Priceless-P:
@patricoferris pls how do I format `json (Ezjsonm.value)`?

patricoferris:
The quick escape-hatch for formatting something is to convert it to a string and use `Format.pp_print_string` :))

Priceless-P:
I can't even pretend I totally understand how formatting works in Ocaml. I do not know where to place this `Format.pp_print_string` but I think I was able to convert `json` to a string with `Ezjsonm.value_to_string json`
So I got rid of the type compatibility errors. However, `dune runtest` results in this error
```
[33mASSERT[0m same arcs
[33mASSERT[0m same foreign_member
[31mFAIL[0m same foreign_member

   Expected: `[32marcs
{"type":"Topology","objects":{"example":{"type":"GeometryCollection","geometries":[{"type":"Point","properties":{"prop0":"value0"},"coordinates":[102,0.5],"arcs":[0]},{"type":"LineString","properties":{"prop0":"value0","prop1":0},"arcs":[0]},{"type":"Polygon","properties":{"prop0":"value0","prop1":{"this":"that"}},"arcs":[[-2]]}]}},"arcs":[[[102,0],[103,1],[104,0],[105,1]],[[100,0],[101,0],[101,1],[100,1],[100,0]]]}[0m'

   Received: `[31m[0m'

Raised at Alcotest_engine__Test.check in file "src/alcotest-engine/test.ml", line 196, characters 4-261
Called from Dune__exe__Test.main in file "test/topojson/test.ml", line 117, characters 6-111
Called from Alcotest_engine__Core.Make.protect_test.(fun) in file "src/alcotest-engine/core.ml", line 180, characters 17-23
Called from Alcotest_engine__Monad.Identity.catch in file "src/alcotest-engine/monad.ml", line 24, characters 31-35
```

It seems I am testing the `foreign_members` twice. I would have checked "src/alcotest-engine/test.ml" pointed to by `Raised at Alcotest_engine__Test.check in file "src/alcotest-engine/test.ml", line 196, characters 4-261` to see what errors are being raised but I can't seem to find it. Please what do I do? 
I'm so sorry for the stress🥺

patricoferris:
> I can't even pretend I totally understand how formatting works in Ocaml

Neither do I ^^" But your solution is just fine, there are a couple of different ways you can express it and note `Fmt` is just a convenience wrapper around `Format`.

```ocaml
Fmt.pf ppf "%s" (Ezjsonm.value_to_string t)
Fmt.string ppf (Ezjsonm.value_to_string t)
Format.pp_print_string ppf (Ezjsonm.value_to_string t)
```

The output from Alcotest is a little confusing, but you aren't testing it twice. It is just saying it is trying to assert some check and then that check fails. But there are two things to note (1) The expected foreign members isn't correct and (2) the foreign members are empty for the actual value.

To help a little bit here's a function that goes to get the foreign members of the first `Point` in the `GeometryCollection` in the `Topology` object. 

```ocaml
let get_foreign_members_in_point (f : Topojson.Topology.t) =
  let open Topojson in
  let objs = List.hd f.objects |> snd in
  match Geometry.geometry objs with
    | Geometry.Collection (point :: _) -> (
      Geometry.foreign_members point
    )
    | _ -> assert false
```

This, with your changes, should return the list `[ "arcs", `A [`Float 0.]]`. Does that make sense?

Priceless-P:
Looking at `pp_position`, I see that it is of type
`Format.formatter -> Topojson.Geometry.Position.t -> unit`

and the `pp_foreign_members` is of type
`Format.formatter -> (string * Ezjsonm.value) list -> unit`
and is influenced by the formatter `pp_ezjsonm`

Also, `position` in ```Alcotest.(check (array (array position))) "same arcs" f.arcs expected_arcs;``` is  of  ```Topojson.Geometry.Position.t testable```  value while my current ```foreign_memebers```  in  ```Alcotest.(check foreign_members)
        "same foreign_member" expected_foreign_members f.foreign_members;``` has value ```string * Ezjsonm.value) list testable```

So I think I should probably change the val of foreign_members to `Topojson.Topology.t -> (string * Ezjsonm.value) list testable`. If my speculation is correct that means that I have to link `foreign_members` to `get_foreign_members_in_point`. 
I tried to do that here
```
let pp_ezjsonm ppf json = Fmt.pf ppf "%s" (Ezjsonm.value_to_string json) in
  let pp_foreign_member ppf (v : (string * get_foreign_members_in_point) list) =
    Fmt.pf ppf "%a" Fmt.(list (pair string pp_ezjsonm)) v in
  let foreign_members = Alcotest.testable pp_foreign_member Stdlib.( = ) in
```
but i get `Unbound type constructor get_foreign_members_in_point` and honestly i don't the `pp_ezjsonm` will format this as it should.

Also I changed `expected_foreign_member` to

````let expected_foreign_members =
    let open Topojson in
    Geometry.foreign_members```


Please how do I resolve this?

File: test/topojson/test.ml [hunk 6]
```diff
@@ -55,12 +55,23 @@ let expected_arcs =
       |];
     |]
 
+let expected_foreign_members =
```
patricoferris: This is not required, the expected foreign members is a JSON value not a function

test/topojson/test.ml
```diff
@@ -98,24 +109,28 @@ let ezjsonm =
...
+(* type pp_set_formatter_out_functions = {
+     foreign_member : Topojson.Geometry.foreign_members -> unit;
+
+   } *)
 let main () =
   let s = read_file "./test_cases/files/exemplar.json" in
   let json = Ezjsonm.value_from_string s in
+
   let pp_ezjsonm ppf json = Fmt.pf ppf "%s" (Ezjsonm.value_to_string json) in
-  let pp_foreign_member ppf (v : (string * Ezjsonm.value) list) =
```
patricoferris: This was correct before. Foreign members have the type `(string * Ezjsonm.value) list`

test/topojson/test.ml
```diff
@@ -98,24 +109,28 @@ let ezjsonm =
...
+  let pp_foreign_member ppf (v : (string * get_foreign_members_in_point) list) =
     Fmt.pf ppf "%a" Fmt.(list (pair string pp_ezjsonm)) v
   in
-
   let foreign_members = Alcotest.testable pp_foreign_member Stdlib.( = ) in
-  let expected_foreign_members = [ ("arcs", json) ] in
```
patricoferris: You will have to write the expected foreign members by hand by looking at the JSON file. What we need here is

```ocaml
let expected_foreign_members = [ ("arcs", `A [ `Float 0. ]) ] in
```

test/topojson/test.ml
```diff
@@ -98,24 +109,28 @@ let ezjsonm =
...
+
   let topojson_obj = Topojson.of_json json in
   match (topojson_obj, Result.map Topojson.topojson topojson_obj) with
   | Ok t, Ok (Topojson.Topology f) ->
       (* Here we check that the arcs defined in the file are the same as the ones
          we hardcoded above *)
       Alcotest.(check (array (array position))) "same arcs" f.arcs expected_arcs;
       Alcotest.(check foreign_members)
-        "same foreign_member" expected_foreign_members f.foreign_members;
+        d "same foreign_member" expected_foreign_members f.foreign_members;
```
patricoferris: Typo?

patricoferris (changes requested):

test/topojson/test.ml
Priceless-P: Yes. Will correct it immediately.

test/topojson/test.ml
Priceless-P: Okay.
Does that mean that `pp_ezjson` is also correct as it is? Because I have made the specified correction but the test is still not passing

patricoferris: > let pp_ezjsonm ppf json = Fmt.pf ppf "%s" (Ezjsonm.value_to_string json) in

That looks good to me yep

test/topojson/test.ml
```diff
@@ -109,28 +97,32 @@ let ezjsonm =
...
   let topojson_obj = Topojson.of_json json in
   match (topojson_obj, Result.map Topojson.topojson topojson_obj) with
   | Ok t, Ok (Topojson.Topology f) ->
       (* Here we check that the arcs defined in the file are the same as the ones
          we hardcoded above *)
-      Alcotest.(check (array (array position))) "same arcs" f.arcs expected_arcs;
       Alcotest.(check foreign_members)
-        d "same foreign_member" expected_foreign_members f.foreign_members;
+        "same foreign_members" expected_foreign_members f.foreign_members;
+      Alcotest.(check (array (array position))) "same arcs" f.arcs expected_arcs;
```
patricoferris, line 123: You need to use the `get_foreign_members_in_point` function

patricoferris (changes requested):

Priceless-P:
>You need to use the `get_foreign_members_in_point` function 

Okay, I was wondering where to bring it in. 

So it would look something like
```
Alcotest.(check foreign_members)
        "same foreign_members" expected_foreign_members
        (get_foreign_members_in_point  f.foreign_members);
```
But then, `get_foreign_members_in_point` takes in a `Topojson.Topology.t` argument. So I tried `(get_foreign_members_in_point  f)`. It gives no error. But I know that application is either wrong or incomplete. Please how can I apply the `get_foreign_members_in_point` to get the desired result?

patricoferris:
Yes, the `(get_foreign_members_in_point f)` is correct and gives me: 

```
ASSERT same foreign_members
FAIL same foreign_members

   Expected: `arcs
[0]'
   Received: `arcs
[0.1]'
```

Which is because I got the wrong value for the points it should be `0.1` and not `0` as I said before

Priceless-P:
>Which is because I got the wrong value for the points it should be 0.1 and not 0 as I said before

I have corrected it and the check passed.

Thank you very much for not giving up on me @patricoferris .

patricoferris:
Yay, thanks 🎉

PR #16, opened 2022-10-20
Provide an example of building a TopoJSON object and writing it to a string

### LINKED ISSUE #8: MDX Docs: Provide an example of building a TopoJSON object and writing it to a string
This requires:

 - Providing a constructor function for making [values of type `t`](https://github.com/geocaml/ocaml-topojson/blob/15b98302c424746a6709a3b948aad9fc765d6b48/src/topojson/topojson_intf.ml#L215)
 - Then building a TopoJSON object
 - Serialise the object to JSON

OmkarAcharekar:
Fixes #8  
 As I understood this issue it requires:

- [x]  Providing a constructor function for making values of type t
- [x]  Then building a TopoJSON object
- [x]   Serialise the object to JSON

* 1 point : It says to create a contructor function to make up values of type t . So I created this function
`let constr (topojson : topojson) (bbox : float array option) : t = { topojson; bbox }`
I think I should add this in [this](https://github.com/geocaml/ocaml-topojson/blob/15b98302c424746a6709a3b948aad9fc765d6b48/src/topojson/topojson_intf.ml#L221) place in the source code.

* 2 point : I guess it is asking to build TopoJson Object from the type t
So  `val topojson : t -> topojson`  is the function that takes a` t `and returns a topojson object.

* 3 point: After getting the object from above function we need to convert it into Json . This can be done by` to_json `function 

@patricoferris Sir , please let me know  for corrections . Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,4 +70,23 @@ A small example how this library is efficient in reading a `json` file _(more pa
 - : (Topojson.t, [ `Msg of string ]) result = Ok <abstr>
 ```
 
+### Example of building a TopoJSON object and writing it to a string
+
+```ocaml
+
+let topojs = {
+    type = "Topology";
+    objects = { example = {
+        type = "GeometryCollection"; geometries = [{
+                type = "Point";properties = {prop0 = "value0"};coordinates = [102; 0.5]}; {type = "LineString";properties = { prop0 = "value0";prop1 = 0}; arcs = [0]
+             } ; {
+                type = "Polygon"; properties = { prop0 = "value0";  prop1 = { this = "that" }}; 
+                arcs = [[-2]] } ]}}; arcs = [[[102; 0];[103; 1];[104; 0];[105; 1]];
+            ]}     
+let bbox = [| 1.2; 2.1; 3.1; |] 
+let type_t =  constr topojs bbox   // calling the contructor 
+let tponJson_object = topojson type_t // converting type_t to tponJson_object
+let str = to_json tponJson_object   //  converting tponJson_object to string
+```
+
 
--- a/src/topojson/topojson_intf.ml
+++ b/src/topojson/topojson_intf.ml
@@ -220,6 +220,10 @@ module type S = sig
   type topojson = Topology of Topology.t | Geometry of Geometry.t
   type t
 
+  let constr (topojson : topojson) (bbox : float array option) : t = { topojson; bbox }
+
+
+
   val topojson : t -> topojson
   val bbox : t -> float array option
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -70,4 +70,23 @@ A small example how this library is efficient in reading a `json` file _(more pa
...
+    type = "Topology";
+    objects = { example = {
+        type = "GeometryCollection"; geometries = [{
+                type = "Point";properties = {prop0 = "value0"};coordinates = [102; 0.5]}; {type = "LineString";properties = { prop0 = "value0";prop1 = 0}; arcs = [0]
+             } ; {
+                type = "Polygon"; properties = { prop0 = "value0";  prop1 = { this = "that" }}; 
+                arcs = [[-2]] } ]}}; arcs = [[[102; 0];[103; 1];[104; 0];[105; 1]];
+            ]}     
+let bbox = [| 1.2; 2.1; 3.1; |] 
+let type_t =  constr topojs bbox   // calling the contructor
```
patricoferris, line 87:
```suggestion
let type_t =  constr topojs bbox
```

Note that OCaml comments are enclosed like `(* my comment *)`

README.md
```diff
@@ -70,4 +70,23 @@ A small example how this library is efficient in reading a `json` file _(more pa
 - : (Topojson.t, [ `Msg of string ]) result = Ok <abstr>
 ```
 
+### Example of building a TopoJSON object and writing it to a string
+
+```ocaml
+
+let topojs = {
+    type = "Topology";
+    objects = { example = {
```
patricoferris, line 79: As it mentioned in the issue it would be better to add functions that allow you to construct a new TopoJSON value. So we end up with something like:

```ocaml
let topojson = 
  let objects = [ (* Create some objects *) ]
  let topology = Topology.v objects arcs in
  Topojson.v topology 
```

patricoferris (changes requested):
Thanks! With respect to your points:

>1 point : It says to create a contructor function to make up values of type t . So I created this function
let constr (topojson : topojson) (bbox : float array option) : t = { topojson; bbox }
I think I should add [this](https://github.com/geocaml/ocaml-topojson/blob/15b98302c424746a6709a3b948aad9fc765d6b48/src/topojson/topojson_intf.ml#L221) in this place in the source code.

Yes that seems correct, I would call it `v` and note you don't need to add the type annotations to the arguments.

> 2 point : I guess it is asking to build TopoJson Object from the type t
So val topojson : t -> topojson is the function that takes atand returns a topojson object.

For constructing an objects we already have everything you need because the type is `(string * Geometry.t) list`. So for example all you need is:

```ocaml
[ "example", Geometry.v ... ]
```

> 3 point: After getting the object from above function we need to convert it into Json . This can be done by `to_json` function

Yep.

patricoferris:
How's this going @OmkarAcharekar, anything I can help with?

patricoferris:
This was completed in #33

PR #18, opened 2022-10-29
add optional transform objects

mariamadronah:
Issue: Add optional transform objects to topology object

### DIFF
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -380,10 +380,10 @@ module Make (J : Intf.Json) = struct
       objects : (string * Geometry.t) list;
       arcs : Geometry.Position.t array array;
       foreign_members : (string * json) list;
-      transform : string * Geometry.Position.t array;
+      transform : transform option;
     }
 
-    type transform = { scale : float * float; translate : float * float }
+    and transform = { scale : float array; translate : float array }
 
     let keys_in_use =
       [
@@ -404,12 +404,18 @@ module Make (J : Intf.Json) = struct
       | Error _ -> []
 
     let base_of_json json =
-      match
-        ( J.find json [ "objects" ],
-          J.find json [ "arcs" ],
-          J.find json [ "transform" ] )
-      with
-      | Some objects, Some arcs, Some transform ->
+      match (J.find json [ "scale" ], J.find json [ "translate" ]) with
+      | Some scale, Some translate ->
+          let* scale = J.to_array (decode_or_err J.to_float) scale in
+
+          let* translate = (J.to_array (decode_or_err J.to_float)) translate in
+
+          Ok { scale; translate }
+      | _, _ -> Error (`Msg "No transform field in Topology object!")
+
+    let base_of_json json =
+      match (J.find json [ "objects" ], J.find json [ "arcs" ]) with
+      | Some objects, Some arcs ->
           let* objects = J.to_obj objects in
           let geometries =
             List.map
@@ -423,39 +429,19 @@ module Make (J : Intf.Json) = struct
                     (decode_or_err (J.to_array (decode_or_err J.to_float)))))
               arcs
           in
-
-          (*let* transform =
-            (*J.to_string transform in
-            let float=
-               (J.to_list(J.to_array(decode_or_err J.to_float)))
-              transform*)
-            J.to_obj in let float= (J.to_list(J.to_array(decode_or_err (J.to_float J.to_float))))
-              transform*)
-          let* transform = J.to_obj transform in
-
-          let float = (J.to_array (decode_or_err J.to_float)) transform in
-
+          let transform = base_of_json json in
           let fm = foreign_members_of_json json in
-          Ok
-            {
-              objects = geometries;
-              arcs;
-              transform = float;
-              foreign_members = fm;
-            }
-      | _, _, _ ->
-          Error
-            (`Msg "No objects, arcs and/or transform field in Topology object!")
+          Ok { objects = geometries; arcs; transform; foreign_members = fm }
+      | _, _ -> Error (`Msg "No objects and/or arcs field in Topology object!")
 
-    let to_json ?bbox { objects; arcs; transform; foreign_members } =
+    let to_json ?bbox { objects; arcs; foreign_members; transform } =
       J.obj
         ([
            ("type", J.string "Topology");
            ( "objects",
              J.obj (List.map (fun (k, v) -> (k, Geometry.to_json v)) objects) );
            ("arcs", J.array (J.array (J.array J.float)) arcs);
-           (*("transform", ( (J.list(J.array J.float)) transform));*)
-           ("transform", J.obj (J.array J.float) transform);
+           ("transform", J.obj [ ("scale", [ 1. ]) ] transform);
          ]
         @ bbox_to_json_or_empty bbox
         @ foreign_members)
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -380,9 +380,10 @@ module Make (J : Intf.Json) = struct
       objects : (string * Geometry.t) list;
       arcs : Geometry.Position.t array array;
       foreign_members : (string * json) list;
-      transform:  (string* Geometry.Position.t array )list;
-       
+      transform : string * Geometry.Position.t array;
     }
+
+    type transform = { scale : float * float; translate : float * float }
 
     let keys_in_use =
       [
@@ -398,13 +399,17 @@ module Make (J : Intf.Json) = struct
 
     let foreign_members_of_json json =
       match J.to_obj json with
-       | Ok assoc ->
+      | Ok assoc ->
           List.filter (fun (k, _v) -> not (List.mem k keys_in_use)) assoc
       | Error _ -> []
 
     let base_of_json json =
-      match (J.find json [ "objects" ], J.find json [ "arcs" ], J.find json [ "transform" ]) with
-      | Some objects, Some arcs, Some transform->
+      match
+        ( J.find json [ "objects" ],
+          J.find json [ "arcs" ],
+          J.find json [ "transform" ] )
+      with
+      | Some objects, Some arcs, Some transform ->
           let* objects = J.to_obj objects in
           let geometries =
             List.map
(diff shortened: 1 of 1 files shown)


src/topojson/topojson.ml
```diff
@@ -380,6 +380,8 @@ module Make (J : Intf.Json) = struct
       objects : (string * Geometry.t) list;
       arcs : Geometry.Position.t array array;
       foreign_members : (string * json) list;
+      transform:  (string* Geometry.Position.t array )list;
```
patricoferris: I think we can more specific here given the specification: 

> A transform must have a member with the name “scale” whose value is a two-element array of numbers.
> A transform must have a member with the name “translate” whose value is a two-element array of numbers.

So I think we can define a new type for the transform object. 

```
type transform = {
  scale : float * float;
  translate : float * float
}
```

Note we'll have to do a conversion by hand from a JSON array of numbers to a pair of OCaml floats. 

Also, it should be optional i.e. `transform : transform option`

src/topojson/topojson.ml
```diff
@@ -396,13 +398,13 @@ module Make (J : Intf.Json) = struct
 
     let foreign_members_of_json json =
       match J.to_obj json with
-      | Ok assoc ->
+       | Ok assoc ->
```
patricoferris: Have a look at the CONTRIBUTING guide for how to format the code using `dune`.

src/topojson/topojson.ml
```diff
@@ -415,18 +417,35 @@ module Make (J : Intf.Json) = struct
...
+          in
+
+          let* transform =
+           (* J.to_list( ("scale", "translate")J.to_array(decode_or_err J.to_float))
+              transform*)
+            (*J.to_list((scale J.to_array(decode_or_err J.to_float)),
+              (translate J.to_array(decode_or_err J.to_float)) )*)
+            
+            J.to_string transform in 
+            let (Geometry.Position.t )=
```
patricoferris: This isn't value OCaml code , let's just use `float` instead of a position.

src/topojson/topojson.ml
```diff
@@ -415,18 +417,35 @@ module Make (J : Intf.Json) = struct
...
 
-    let to_json ?bbox { objects; arcs; foreign_members } =
+    let to_json ?bbox { objects; arcs; foreign_members; transform } =
       J.obj
         ([
            ("type", J.string "Topology");
            ( "objects",
              J.obj (List.map (fun (k, v) -> (k, Geometry.to_json v)) objects) );
            ("arcs", J.array (J.array (J.array J.float)) arcs);
+           ("transform", J.string ("scale", "translate") (J.list(J.array J.float) transform));
```
patricoferris: You will need to use the `J.obj` function

patricoferris (changes requested):
Thanks! This is heading in the right direction, just a few comments and pointers to help you get it working

mariamadronah:
hello @patricoferris , I have made some new changes. Please help review. Thanks

src/topojson/topojson.ml
```diff
@@ -380,10 +380,11 @@ module Make (J : Intf.Json) = struct
       objects : (string * Geometry.t) list;
       arcs : Geometry.Position.t array array;
       foreign_members : (string * json) list;
-      transform:  (string* Geometry.Position.t array )list;
-       
+      transform : string * Geometry.Position.t array;
     }
 
+    type transform = { scale : float * float; translate : float * float }
```
```
      transform : transform option;
     }

     and transform = { scale : float * float; translate : float * float }
```

src/topojson/topojson.ml
```diff
@@ -416,17 +423,39 @@ module Make (J : Intf.Json) = struct
...
+          (*let* transform =
+            (*J.to_string transform in
+            let float=
+               (J.to_list(J.to_array(decode_or_err J.to_float)))
+              transform*)
+            J.to_obj in let float= (J.to_list(J.to_array(decode_or_err (J.to_float J.to_float))))
+              transform*)
+          let* transform = J.to_obj transform in
+
+          let float = (J.to_array (decode_or_err J.to_float)) transform in
```
patricoferris: `transform` is an object here not a float array, you will need to pull out the two float arrays under the two keys. `float` is also not a great name for this variable.

src/topojson/topojson.ml
```diff
@@ -416,17 +423,39 @@ module Make (J : Intf.Json) = struct
...
-    let to_json ?bbox { objects; arcs; foreign_members } =
+    let to_json ?bbox { objects; arcs; transform; foreign_members } =
       J.obj
         ([
            ("type", J.string "Topology");
            ( "objects",
              J.obj (List.map (fun (k, v) -> (k, Geometry.to_json v)) objects) );
            ("arcs", J.array (J.array (J.array J.float)) arcs);
+           (*("transform", ( (J.list(J.array J.float)) transform));*)
+           ("transform", J.obj (J.array J.float) transform);
```
patricoferris: You'll need to convert `transform` to a `(string * json) list` encoding for example:

```json
{ "someKey": [ 1.0 ] }
```

Would be 

``ocaml
[ "someKey", `A [ `Float 1.0 ]]
```

**Review by patricoferris — CHANGES_REQUESTED** [review]

File: src/topojson/topojson.ml (hunk 6 above)
Comment (mariamadronah) [reply src/topojson/topojson.ml]: @patricoferris can this work to pull out the two float arrays under the two keys??
let* transform = 
            match {} with
            |  {scale; translate} -> string_of_float scale ^" '" ^ string_of_float translate;
            transform

Comment (patricoferris) [reply src/topojson/topojson.ml]: No, there's an example just above of accessing these kinds of values i.e. json values associated with some key.

```ocaml
 match
  (J.find json [ "objects" ],
   J.find json [ "arcs" ],
   J.find json [ "transform" ] )
       with
       | Some objects, Some arcs, Some transform ->
```

src/topojson/topojson.ml
```diff
@@ -380,10 +380,10 @@ module Make (J : Intf.Json) = struct
       objects : (string * Geometry.t) list;
       arcs : Geometry.Position.t array array;
       foreign_members : (string * json) list;
-      transform : string * Geometry.Position.t array;
+      transform : transform option;
     }
 
-    type transform = { scale : float * float; translate : float * float }
+    and transform = { scale : float array; translate : float array }
```
patricoferris, line 386: This should be `float * float` and not an array -- JSON doesn't have a way of expressing pairs and that's why the TopoJSON spec uses an array, but we can convert to a pair.

src/topojson/topojson.ml
```diff
@@ -404,12 +404,18 @@ module Make (J : Intf.Json) = struct
       | Error _ -> []
 
     let base_of_json json =
-      match
-        ( J.find json [ "objects" ],
-          J.find json [ "arcs" ],
-          J.find json [ "transform" ] )
-      with
-      | Some objects, Some arcs, Some transform ->
+      match (J.find json [ "scale" ], J.find json [ "translate" ]) with
```
patricoferris, line 407: Trying to find the keys `scale` and `translate` need to be performed on the `transform` object not the `json` one.

src/topojson/topojson.ml
```diff
@@ -404,12 +404,18 @@ module Make (J : Intf.Json) = struct
...
+      | Some scale, Some translate ->
+          let* scale = J.to_array (decode_or_err J.to_float) scale in
+
+          let* translate = (J.to_array (decode_or_err J.to_float)) translate in
+
+          Ok { scale; translate }
+      | _, _ -> Error (`Msg "No transform field in Topology object!")
```
patricoferris, line 414: This is not an error, the transform object is optional.

src/topojson/topojson.ml
```diff
@@ -404,12 +404,18 @@ module Make (J : Intf.Json) = struct
       | Error _ -> []
 
     let base_of_json json =
```
patricoferris, line 406: Let's rename this to `transform_of_json`

src/topojson/topojson.ml
```diff
@@ -423,39 +429,19 @@ module Make (J : Intf.Json) = struct
...
+    let to_json ?bbox { objects; arcs; foreign_members; transform } =
       J.obj
         ([
            ("type", J.string "Topology");
            ( "objects",
              J.obj (List.map (fun (k, v) -> (k, Geometry.to_json v)) objects) );
            ("arcs", J.array (J.array (J.array J.float)) arcs);
-           (*("transform", ( (J.list(J.array J.float)) transform));*)
-           ("transform", J.obj (J.array J.float) transform);
+           ("transform", J.obj [ ("scale", [ 1. ]) ] transform);
```
patricoferris, line 444: You will probably need a function that converts the OCaml transform record back into a JSON value, just like you have one for converting from JSON. Converting the pair for scale would look like

```OCaml
J.obj [ "scale", J.array J.float [| fst t.scale; snd t.scale |]
```

But note, the transform object is optional so if it is `None` we don't want to add anything.

patricoferris (changes requested):

src/topojson/topojson.ml
Priceless-P: Okay , so what should 
```ocaml
_ , _  ->
``` 
return?

patricoferris: The error is ok actually but it is being applied incorrectly. It is trying to read the fields of the transform object so if scale and translate are not present it is indeed an error (but the message is wrong). Where this function (confusingly named `base_of_json`) is used is not reading the correct JSON which is why it isn't right.

Priceless-P: Okay. So I need to rename it to `tranform_of_json` as you said in the earlier comment below. But I'm still not clear on what the function is to do.

I saw [here](https://github.com/topojson/topojson-client#transform)
> [#](https://github.com/topojson/topojson-client#transform) topojson.transform(transform) [<>](https://github.com/topojson/topojson-client/blob/master/src/transform.js)
If the specified [transform object](https://github.com/topojson/topojson-specification/blob/master/README.md#212-transforms) is non-null, returns a [point transform function](https://github.com/topojson/topojson-client#_transform) to remove delta-encoding and apply the transform. If the transform is null, returns the identity function.

And [point transform function](https://github.com/topojson/topojson-client#_transform) says 
>[#](https://github.com/topojson/topojson-client#_transform) transform(point[, index])
Applies this transform function to the specified point, returning a new point with the transformed coordinates. If the specified index is truthy, the input point is treated as relative to the previous point passed to this transform, as is the case with delta-encoded arcs

Is this in line with what `tranform_of_json` is supposed to do?

patricoferris: I wouldn't look at the topojson client right now -- that's additional functionality provided by a library. The specification is what we need https://github.com/topojson/topojson-specification#212-transforms

Priceless-P: Okay

patricoferris:
Closing this in favour of #32

Issue #17, opened 2022-10-25
Test that the `exemplar.json` GeometryCollection is parsed correctly


patricoferris:
Currently we don't actually test if the parsing of the TopoJSON document is correct with respect to the inner GeometryCollection i.e. the values:

```json 
{
    "type": "Point",
    "properties": {
      "prop0": "value0"
    },
    "coordinates": [102, 0.5]
  },
  {
    "type": "LineString",
    "properties": {
      "prop0": "value0",
      "prop1": 0
    },
    "arcs": [0]
  },
  {
    "type": "Polygon",
    "properties": {
      "prop0": "value0",
      "prop1": {
        "this": "that"
      }
    },
    "arcs": [[-2]]
  }
```

I think the test should try to extract these values from the OCaml value and check some of them for example that the first element is indeed a point and it contains the coordinates `[102, 0.5]`.

olybeke:
Hi mentor @patricoferris, I'm Oluchi Nwankwo, an outreachy applicant. I would like to work on this issue. can you assign it to me?

patricoferris:
Hey @olybeke, have got your OCaml development environment all configured? If so, then sure :))

olybeke:
Yes, I have

patricoferris:
Great!

Priceless-P:
Hi @patricoferris, please I'd like to fix this issue.

patricoferris:
Awesome, let me know if you need help :))

Priceless-P:
Okay, sure

patricoferris:
We have a bunch of tests which cover all of this I think

PR #25, opened 2022-11-03
testing parsing for exemplar.json

olybeke:
issue #17: test that the `exemplar.json` GeomentryCollection is parsed correctly.

Mentor @patricoferris  I tried testing one of the Geometry objects `multipolygon` to see if it's parsed and has coordinates, but unfortunately after running a test on it using `dune runtest` it showed failed and I wonder what could be wrong.

Although, I'm still having a little misunderstanding on how to go about this. I would appreciate some help or pointers on how to resolve this issue.

### DIFF
--- a/test/topojson/test_cases/files/exemplar.json
+++ b/test/topojson/test_cases/files/exemplar.json
@@ -28,6 +28,16 @@
             }
           },
           "arcs": [[-2]]
+        },
+        {
+          "type":"multipolygon",
+          "properties": {
+            "prop0": "value0",
+            "prop1": {
+              "this":"that"
+            }
+          },
+          "arcs":[[-1]]
         }
       ]
     }


patricoferris:
Two reasons this won't work:

1. I've interpreted the GeoJSON spec as saying "A GeoJSON object has a member with the name "type".  The value of
      the member MUST be one of the GeoJSON types." to mean they must be _exact_, so it should be `"type": "MultiPolygon"`.
2. A multipolygon is multiple polygons. A polygon is multiple line segments. So arcs should be multiple, multiple arc index lists. Concretely it only makes sense to have at least `"arcs":[[[-1]]]` although we should actually check that properly.

olybeke:
Alright, I will work on and also carry of the test.
Thank you.

PR #28, opened 2022-12-09
Accessor functions

Priceless-P:
Functions to access topojson geometries (Points, LineString, Polgon and MultiPolygon)

### DIFF
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -60,8 +60,9 @@ module Make (J : Intf.Json) = struct
       | None, _ ->
           Error
             (`Msg
-              ("JSON should" ^ "have a key-value for `type' whilst parsing "
-             ^ typ))
+              ("JSON should"
+              ^ "have a key-value for `type' whilst parsing "
+              ^ typ))
       | _, None -> Error (`Msg "JSON should have a key-value for `coordinates'")
       | Some typ, Some coords -> (
           let* typ = J.to_string typ in
@@ -74,8 +75,9 @@ module Make (J : Intf.Json) = struct
       | None, _ ->
           Error
             (`Msg
-              ("JSON should" ^ "have a key-value for `type' whilst parsing "
-             ^ typ))
+              ("JSON should"
+              ^ "have a key-value for `type' whilst parsing "
+              ^ typ))
       | _, None -> Error (`Msg "JSON should have a key-value for `arcs'")
       | Some typ, Some arcs -> (
           let* typ = J.to_string typ in
@@ -131,7 +133,8 @@ module Make (J : Intf.Json) = struct
              ("type", J.string typ); ("coordinates", Position.to_json position);
            ]
           @ properties_or_null properties
-          @ bbox_to_json_or_empty bbox @ foreign_members)
+          @ bbox_to_json_or_empty bbox
+          @ foreign_members)
     end
 
     module MultiPoint = struct
@@ -155,7 +158,8 @@ module Make (J : Intf.Json) = struct
              ("coordinates", J.array Position.to_json positions);
            ]
           @ properties_or_null properties
-          @ bbox_to_json_or_empty bbox @ foreign_members)
+          @ bbox_to_json_or_empty bbox
+          @ foreign_members)
     end
 
     module LineString = struct
@@ -170,7 +174,8 @@ module Make (J : Intf.Json) = struct
         J.obj
           ([ ("type", J.string typ); ("arcs", Arcs.to_json arc) ]
           @ properties_or_null properties
-          @ bbox_to_json_or_empty bbox @ foreign_members)
+          @ bbox_to_json_or_empty bbox
+          @ foreign_members)
     end
 
     module MultiLineString = struct
@@ -189,7 +194,8 @@ module Make (J : Intf.Json) = struct
         J.obj
           ([ ("type", J.string typ); ("arcs", J.array Arcs.to_json arcs) ]
           @ properties_or_null properties
-          @ bbox_to_json_or_empty bbox @ foreign_members)
+          @ bbox_to_json_or_empty bbox
+          @ foreign_members)
     end
 
     module Polygon = struct
@@ -212,7 +218,8 @@ module Make (J : Intf.Json) = struct
         J.obj
           ([ ("type", J.string typ); ("arcs", J.array (J.array J.int) arcs) ]
           @ properties_or_null properties
-          @ bbox_to_json_or_empty bbox @ foreign_members)
+          @ bbox_to_json_or_empty bbox
+          @ foreign_members)
     end
 
     module MultiPolygon = struct
@@ -235,7 +242,8 @@ module Make (J : Intf.Json) = struct
              ("arcs", J.array (J.array (J.array J.int)) arcs);
            ]
           @ properties_or_null properties
-          @ bbox_to_json_or_empty bbox @ foreign_members)
+          @ bbox_to_json_or_empty bbox
+          @ foreign_members)
     end
 
     type geometry =
@@ -432,7 +440,8 @@ module Make (J : Intf.Json) = struct
                ("geometries", J.list to_json c);
              ]
             @ properties_or_null t.properties
-            @ bbox_to_json_or_empty bbox @ t.foreign_members)
+            @ bbox_to_json_or_empty bbox
+            @ t.foreign_members)
   end
 
   module Topology = struct
@@ -490,7 +499,8 @@ module Make (J : Intf.Json) = struct
              J.obj (List.map (fun (k, v) -> (k, Geometry.to_json v)) objects) );
            ("arcs", J.array (J.array (J.array J.float)) arcs);
          ]
-        @ bbox_to_json_or_empty bbox @ foreign_members)
+        @ bbox_to_json_or_empty bbox
+        @ foreign_members)
   end
 
   type topojson = Topology of Topology.t | Geometry of Geometry.t
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -112,12 +112,6 @@ module Make (J : Intf.Json) = struct
       let to_json arr = J.array J.float arr
     end
 
-    module Arcs = struct
-      type t = int array
-
-      let to_json arr = J.array J.int arr
-    end
-
     module Point = struct
       type t = Position.t
 
@@ -136,7 +130,12 @@ module Make (J : Intf.Json) = struct
           @ bbox_to_json_or_empty bbox
           @ foreign_members)
     end
+    module Arcs = struct
+      type t = int array | float array
+      let v t = t
 
+      let to_json arr = J.array J.int arr
+    end
     module MultiPoint = struct
       type t = Point.t array
 
--- a/src/topojson/topojson.ml
+++ b/src/topojson/topojson.ml
@@ -60,9 +60,8 @@ module Make (J : Intf.Json) = struct
       | None, _ ->
           Error
             (`Msg
-              ("JSON should"
-              ^ "have a key-value for `type' whilst parsing "
-              ^ typ))
+              ("JSON should" ^ "have a key-value for `type' whilst parsing "
+             ^ typ))
       | _, None -> Error (`Msg "JSON should have a key-value for `coordinates'")
       | Some typ, Some coords -> (
           let* typ = J.to_string typ in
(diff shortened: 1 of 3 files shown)


src/topojson/topojson.ml
```diff
@@ -242,8 +235,33 @@ module Make (J : Intf.Json) = struct
              ("arcs", J.array (J.array (J.array J.int)) arcs);
            ]
           @ properties_or_null properties
-          @ bbox_to_json_or_empty bbox
-          @ foreign_members)
+          @ bbox_to_json_or_empty bbox @ foreign_members)
+    end
+
+    module Accessor = struct
```
patricoferris: I think it would be more convenient if we didn't nest these inside an `Accessor` module. I think I did that for GeoJSON but probably shouldn't have.

src/topojson/topojson.ml
```diff
@@ -242,8 +235,33 @@ module Make (J : Intf.Json) = struct
...
+
+      let get_point = function
+        | Topojson.Geometry.Point p -> Ok p
+        | _ -> Error (`Msg "Expected point")
+
+      let get_point_exn = function
+        | Topojson.Geometry.Point p -> p
+        | _ -> invalid_arg "Expected point"
+
+      let get_point = function
```
patricoferris: You need to change these from points to the other possible geometries such as multipolygon.

src/topojson/topojson_intf.ml
```diff
@@ -241,3 +241,29 @@ module type Topojson = sig
       parser and constructor. *)
   module Make (J : Json) : S with type json = J.t
 end
+
+module type Accessor = sig
```
patricoferris: Same point as above, let's remove the wrapping `Accessor` module so it is more direct to use these functions

src/topojson/topojson_intf.ml
```diff
@@ -241,3 +241,29 @@ module type Topojson = sig
...
+    Geometry.geometry -> (Geometry.LineString.t, [> `Msg of string ]) result
+
+  val get_linestring_exn : Geometry.geometry -> Geometry.LineString.t
+
+  val get_polygon :
+    Geometry.geometry -> (Geometry.Polygon.t, [> `Msg of string ]) result
+
+  val get_polygon_exn : Geometry.geometry -> Geometry.Polygon.t
+
+  val get_multipolygon :
```
patricoferris: Looking good, let's add _all_ of the geometries! :))

patricoferris (changes requested):
Thanks! This is looking great, just some points!

src/topojson/topojson_intf.ml
Priceless-P: If i remove 

```ocaml
module type Accessor = sig
  type t
```

I get this error ```Value declarations are only allowed in signatures``` for these parts of the code
```ocaml
val get_point :
    Geometry.geometry -> (Geometry.Point.t, [> `Msg of string ]) result
  val get_point_exn : Geometry.geometry -> Geometry.Point.t
```

src/topojson/topojson.ml
Priceless-P: Okay.

But this part 
```ocaml 
Topojson.Geometry.Point
```  
also gives error `Unbound module Topojson`

src/topojson/topojson_intf.ml
Priceless-P: @patricoferris Okay

src/topojson/topojson.ml
patricoferris: Because you are in the topojson file, you are in the `Topojson` module so you need to remove that bit.

src/topojson/topojson_intf.ml
patricoferris: Ah I didn't see they weren't with the Geometry module type, if you move them up there that would be better. You won't need to redeclare `Geoemetry` either so your functions become:

```ocaml
val get_point : geometry -> (Geometry.Point.t, [> `Msg of string ]) result

  val get_point_exn : geometry -> Point.t
```

src/topojson/topojson_intf.ml
```diff
@@ -181,6 +200,12 @@ module type Geometry = sig
   val geometry : t -> geometry
   (** [geometry t] returns the geometry associated with the object. *)
 
+  (* val geometry_v :
+     ?properties:properties ->
+     geometry ->
+     t *)
+  val geometry_to_json : t -> json
```
patricoferris: This is unnecessary as it is included by

```ocaml
include Json_conv with type t := t and type json := json
```

test/topojson/test.ml
```diff
@@ -105,13 +105,60 @@ let get_foreign_members_in_point (f : Topojson.Topology.t) =
   | _ -> assert false
 
 let pp_ezjsonm ppf json = Fmt.pf ppf "%s" (Ezjsonm.value_to_string json)
+let tt = Topojson.Geometry.geometry
```
patricoferris: Are these three lines needed anymore ?

test/topojson/test.ml
```diff
@@ -105,13 +105,60 @@ let get_foreign_members_in_point (f : Topojson.Topology.t) =
...
+(* let v = Topojson.Geometry.geometry_to_json in
+   Geometry.to_json v *)
+
+let pp ppf tt =
+  Fmt.pf ppf "%s"
+    (Ezjsonm.value_to_string (Topojson.Geometry.geometry_to_json tt))
```
patricoferris: I think this just needs to be

```suggestion
    (Ezjsonm.value_to_string (Topojson.Geometry.to_json tt))
```

test/topojson/test.ml
```diff
@@ -105,13 +105,60 @@ let get_foreign_members_in_point (f : Topojson.Topology.t) =
   | _ -> assert false
 
 let pp_ezjsonm ppf json = Fmt.pf ppf "%s" (Ezjsonm.value_to_string json)
+let tt = Topojson.Geometry.geometry
+(* let v = Topojson.Geometry.geometry_to_json in
+   Geometry.to_json v *)
+
+let pp ppf tt =
```
patricoferris:
```suggestion
let pp_geometry ppf tt =
```

test/topojson/test.ml
```diff
@@ -105,13 +105,60 @@ let get_foreign_members_in_point (f : Topojson.Topology.t) =
...
+  Fmt.pf ppf "%s"
+    (Ezjsonm.value_to_string (Topojson.Geometry.geometry_to_json tt))
+
+let pp_coord ppf (v : Topojson.Geometry.geometry list) =
```
patricoferris: The names should probably be tied to what they pretty print?

```suggestion
let pp_geometry_list ppf (v : Topojson.Geometry.geometry list) =
```

test/topojson/test.ml
```diff
@@ -105,13 +105,60 @@ let get_foreign_members_in_point (f : Topojson.Topology.t) =
...
+  let open Topojson in
+  let objs = List.hd f.objects |> snd in
+  let s = read_file "./test_cases/files/exemplar.json" in
+  let json = Ezjsonm.value_from_string s in
+  let topo = Topojson.of_json json in
+  match topo with
+  | Ok v -> (
+      match Geometry.geometry objs with
+      | Collection [ point; linestring; polygon; multipolygon ] ->
+          let geo_point = [ Geometry.geometry point ] in
```
patricoferris: Why is everything wrapped in a singleton list? Does it have to be? 

```suggestion
          let geo_point = Geometry.geometry point in
```

patricoferris (changes requested):
Looking good, nearly there, just some comments to help get the code compiling (see the errors in CI)

test/topojson/test.ml
Priceless-P: Removing line 105 causes error `pattern matching not exhausive`
Line 107 is needed in `pp_foreign_member`.
 Let me remove the other two lines

test/topojson/test.ml
Priceless-P: I thought about doing that. But i get error ```
Unbound value Topojson.Geometry.to_json```

test/topojson/test.ml
Priceless-P: I thought to wrapped them because it resolved this error i was getting
```ocaml
This expression has type Geometry.geometry
but an expression was expected of type Geometry.geometry list
```

test/topojson/test.ml
patricoferris: Sorry the viewer on Github is not showing what I wanted, I meant lines 108, 109, 110

test/topojson/test.ml
patricoferris: Ah, https://github.com/geocaml/ocaml-topojson/pull/29 should fix this! Will merge once CI is happy.

test/topojson/test.ml
patricoferris: I think that's because your pp function is trying to print lists

```Ocaml
Fmt.pf ppf "%a" Fmt.(list pp) v
```

The problem is with that function.

Priceless-P:
```ocaml
 [Geometry.geometry objs] ;
``` 
is giving errors ```This variant expression is expected to have type Geometry.geometry
There is no constructor :: within type Topojson.Geometry.geometry``` 
in
```ocaml
| Collection [ point; linestring; polygon; multipolygon ] ->
          let geo_point = Geometry.geometry point in
          Alcotest.(check coordinates)
            "same point" geo_point
            [Geometry.geometry objs] ;
```

I thought about changing it to just `[objs]` but it still gives another error
 ```ocaml
This expression has type Geometry.t but an expression was expected of type
  Geometry.geometry 
```
 I don't think i am getting the expected objects correctly.

**Comment by patricoferris** [comment]
Ah, the problem is we're not testing a list of geometries but each individual element. You are trying to pass a list to a function which expects a `Geometry.geometry` hence the compiler message. Changing it to `[ objs ]` won't help as that is still a list, and what's more a list of the wrong type.

What we really want to do is:

 - Now that we have the point (geo_point) extracted from the JSON we want to test it against some value we expect it to be.
 - We need to construct this `expected` value.
 - Our call to `Alcotest.check` will then become `Alcotest.(check coordinates) "same point" expected geo_point`. Note the naming of the testable value as `coordinates` is not particularly helpful, I suggest changing it to whatever the testable actually is (here the inner geometry type).

If that's clear, if given you a little bit of help in https://github.com/geocaml/ocaml-topojson/commit/7e6227a62c6f857cbb1157e6ea94cdebb7089f2a for the case with the point. Please do ask any questions etc. but I thought that might help clear things up :))

**Comment by Priceless-P** [comment]
Hi mentor @patricoferris, your last comment was really  helpful especially this commit [7e6227a](https://github.com/geocaml/ocaml-topojson/commit/7e6227a62c6f857cbb1157e6ea94cdebb7089f2a). Thank you very much.

I still have one more question. I tried passing the expected arcs for linestring  and polygons

```ocaml
let expected_linestring =
        Geometry.(LineString (LineString.v @@  Arcs.v [ 0 ]))
```

```ocaml
let expected_polygon =
        Geometry.(Polygon (Polygon.v @@ LineString.v [ [ -2 ] ]))
```

But I get error `Unbound value Arc.v` and for expected_polygon I get 
```ocaml
This expression has type LineString.t but an expression was expected of type
  LineString.t array
This expression has type 'a list but an expression was expected of type
  Arcs.t
```

Please how can I pass these values correctly?

patricoferris:
Ah you'll have to modify the interface in the `topojson_intf` file to expose the function to build and `Arcs` value. Currently it only defines the `type t` and you will need also a `val v ... -> t` function.

Priceless-P:
Hi mentor, please review my changes. `Polygon` is  still not working. But I got rid of the error in `LineString` Also I tried to make `Arc.t` to be either an int array or a float array so i can check it against the arcs in the `examplar.json`

patricoferris:
Hmm we need to distinguish between arc indexes which are int arrays and the arcs data which is floating point numbers. The `Arcs` we have are really are indexes. I don't think we need to provide an explicit module and type for them, so instead I think we should remove `Arc` and wherever we have `Arc.t` it should be replaced with an appropriate int array. Does that make sense?

Priceless-P:
@patricoferris Like so ?

src/topojson/topojson_intf.ml
```diff
@@ -105,7 +105,7 @@ module type Geometry = sig
     type t
     (** Arcs is an array of arc indexes *)
 
-    val v : Point.t -> t
+    val v : float array -> t
```
patricoferris: Should this be `int array`

Priceless-P:
Yes. That's true. No errors anymore💃🏻

`dune test` passed too. I changed the value of `linestring arcs` to something else and it failed. That means it works right? Is it satisfactory now?

patricoferris (approved):

patricoferris:
Awesome! Thanks, merged :))

PR #32, opened 2022-12-25
Add optional transform objects to topology object

Priceless-P:


Priceless-P:
@patricoferris, please take a look when you are available :))

src/topojson/topojson.ml
```diff
@@ -479,13 +482,22 @@ module Make (J : Intf.Json) = struct
           List.filter (fun (k, _v) -> not (List.mem k keys_in_use)) assoc
       | Error _ -> []
 
-    let of_json json =
+    let transform_of_json json =
```
patricoferris, line 493: The implementation here doesn't look quite right. I think we will need to use `J.find` to get at the transform object. As well, when decoding the scale and translate values we can convert them to arrays (that's what they are in JSON because JSON doesn't have the notion of pairs). Something like 


```ocaml
let transform_of_json json =
      match J.find json [ "transform "] with
      | None -> None
      | Some transform_object ->
        let scale = decode_or_err (J.to_array (decode_or_err J.to_float)) transform_object in
        let translate = decode_or_err (J.to_array (decode_or_err J.to_float)) transform_object in
        Some { scale = scale.(0), scale.(1); translate = translate.(0), translate.(1) }
```

Does that make sense, note as well how it now returns an optional value. Previously you had `{}` which isn't valid OCaml, but I think I get what you wanted to do, in OCaml we can represent that "nothingness" with `None`.

src/topojson/topojson.ml
```diff
@@ -495,17 +507,19 @@ module Make (J : Intf.Json) = struct
...
 
-    let to_json ?bbox { objects; arcs; foreign_members } =
+    let to_json ?bbox { objects; arcs; foreign_members; transform } =
       J.obj
         ([
            ("type", J.string "Topology");
            ( "objects",
              J.obj (List.map (fun (k, v) -> (k, Geometry.to_json v)) objects) );
            ("arcs", J.array (J.array (J.array J.float)) arcs);
+           ("transform", J.obj [ "scale", J.array J.float [| fst t.scale; snd t.scale |]] );
```
```suggestion
           ("transform", J.obj [ "scale", J.array J.float [| fst transform.scale; snd transform.scale |]] );
```

patricoferris: Note as well that the transform is optional so we need to add it to the list if it exists otherwise we don't.

patricoferris (commented):
Looking good :)) Just need to get the parsing of the transform object working. I think it would also be helpful to add a test for this code too

Priceless-P:
```ocaml
( "scale",
                   J.array J.float
                     [| fst transform.scale; snd transform.scale |] );
```
I get this error here

```ocaml
This expression has type transform option
which is not a record type.
```
Also, I'm not sure I know how to do this

>Note as well that the transform is optional so we need to add it to the list if it exists otherwise we don't.

Priceless-P:
Please how do I make `expected_transform` be of `transform option` instead of just `transform`?

Also here, 
```ocaml
 Alcotest.(check transform_)
        "same transform" expected_transform f.transform;
```
`f.transform` gives error 
```ocaml
This expression has type Topojson.Topology.t
There is no field transform within type Topojson.Topology.t
```
even though it exists and `dune build` is not able to fix.

patricoferris:
> Also, I'm not sure I know how to do this

This should help fix it, you want to great a singleton list if the transform is `Some transform` or the empty list if it is not so you can do

```OCaml
some_list @ (match transform with Some t -> [ ... ] | None -> []) @ potentially_other_things
```

patricoferris:
> f.transform gives error

It probably needs added here:

https://github.com/Priceless-P/ocaml-topojson/blob/cbadc30ccc2f1bef20cc271cc7f5dbc09746ce73/src/topojson/topojson_intf.ml#L252-L256

Priceless-P:
@patricoferris Is `transform_of_json` satisfactory now? 

Also why am i not able to reference ```transform = { scale : float * float; translate : float * float }``` in
```ocaml
let pp_transform ppf = Fmt.pf ppf "%a" transform
```
But it works fine here
```ocaml
let pp_expected_transform ppf (v : transform option) =
  Fmt.pf ppf "%a" Fmt.(pp_transform) v
```

patricoferris:
> Also why am i not able to reference transform = { scale : float * float; translate : float * float } in
> let pp_transform ppf = Fmt.pf ppf "%a" transform

I'm not entirely sure what you mean. In the first example you have `transform` which is presumably unbound to any value and in the second you have `pp_transform` which is different?

Priceless-P:
I meant here `(v: transform option)`

patricoferris:
In the first piece of code `transform` is used as a value, because you have used `"%a"` the value is expected to be of type `Format.formatter -> 'a -> unit` (which is the same as `'a Fmt.t`). In the second piece of code `transform` is used as a _type annotation_.

Priceless-P:
Okay. It's clearer now. Thank you.

However, I am still having a little issue with formatting the `transform` object 🥺

Priceless-P:
Please how do I proceed with this?

PR #33, opened 2022-12-29
Example of building a TopoJSON object and writing it to a string

Priceless-P:


Priceless-P:
@patricoferris please the `Object` okay? Should I proceed with writing it to a string?

Priceless-P:
@patricoferris please the `Object` okay? Should I proceed with writing it to a string?

src/topojson/topojson_intf.ml
```diff
@@ -264,7 +264,7 @@ module type S = sig
 
   val topojson : t -> topojson
   val bbox : t -> float array option
-
+  val vals_of_t : topojson -> topojson -> float array option -> t
```
patricoferris: Ah yes we need to expose constructors for everything in order to build a new value. For the main topojson object we will need:

```ocaml
val v : ?bbox:float array -> topojson -> t
``` 

And then we will need constructors for building `topojson` values too.

README.md
```diff
@@ -90,10 +90,32 @@ A small example how this library is efficient in reading a `json` file _(more pa
...
-# Topojson.of_json ( Ezjsonm.value_from_string topojson_string);; 
+# Topojson.of_json ( Ezjsonm.value_from_string topojson_string);;
 - : (Topojson.t, [ `Msg of string ]) result = Ok <abstr>
 ```
 
+### Example of building a TopoJSON object and writing it to a string
+
+```ocaml
+let topojson =
+  let objects = [ { "example": "type" = "GeometryCollection"; geometries = [{
```
patricoferris: We will need to use functions to construct these values. At the moment this is a mix of JSON/OCaml which will not compile, see my other comment.

patricoferris (commented):
Thanks! We haven't exposed/implemented enough functions to construct values to do this properly yet.

Priceless-P:
@patricoferris please check again. How do I proceed from here?

patricoferris:
Awesome! Great work, to help get this over the line, I hopefully got it to a working state in this commit https://github.com/geocaml/ocaml-topojson/commit/4765b65f618789681171c83481a068726814ccef -- feel free to incorporate that work if it is helpful and please ask any questions you might have :))

Priceless-P:
> feel free to incorporate that work if it is helpful 

Done, thank you.

So we add the signatures to the markup file too. I want to confirm what this `= <abstr>` means in
```ocaml
val t : Topojson.t = <abstr>
```

patricoferris:
We just need a `dune build @fmt` to get the tests all green :))

```ocaml
val t : Topojson.t = <abstr>
```

This means that the type is abstract and that the top-level (which MDX uses) doesn't have a registered printer function to show the actual value. That's completely fine.

Priceless-P:
>This means that the type is abstract and that the top-level (which MDX uses) doesn't have a registered printer function to show the actual value. That's completely fine.

Okay. Thanks for explaining that. I have formatted the code too.

patricoferris:
Great thanks!

PR #35, opened 2023-01-04
Example usage of foreign_members

Priceless-P:
Using `foreign_members` function to read some key-value pairs that are not a part of the specification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,4 +125,5 @@ Given the Topojson Object above, we can extract the foreign members by using the
 
 ```ocaml
 # Topojson.Geometry.foreign_members topojson;;
+- : (string * Topojson.json) list = [("foreign", `String "8")]
 ```
--- a/README.md
+++ b/README.md
@@ -118,3 +118,11 @@ val t : Topojson.t = <abstr>
 - : string =
 "{\"type\":\"Topology\",\"objects\":{\"example\":{\"type\":\"LineString\",\"arcs\":[0],\"foreign\":\"8\"}},\"arcs\":[[[0,0]]]}"
 ```
+
+#### Example usage of foreign_members
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -118,3 +118,11 @@ val t : Topojson.t = <abstr>
 - : string =
 "{\"type\":\"Topology\",\"objects\":{\"example\":{\"type\":\"LineString\",\"arcs\":[0],\"foreign\":\"8\"}},\"arcs\":[[[0,0]]]}"
 ```
+
+#### Example usage of foreign_members
+
+Given the Topojson Object above, we can extract the foreign members by using the foreign_member function in the Topojson.Geometry module.
+
+```ocaml
+let foreign_member_usage = Topojson.Geometry.foreign_members topojson
```
patricoferris: This is great thanks, but I think it would be better to use MDX's top-level syntax so it prints the value too:

```suggestion
# Topojson.Geometry.foreign_members topojson;;
```

Priceless-P: Okay. Done

patricoferris:
You will need to run the code as `dune runtest` will want to promote the result.

Priceless-P:
It seems to work.

patricoferris:
You will need to run `dune runtest` and `dune promote` locally in order to accept the changes -- see the failing tests https://github.com/geocaml/ocaml-topojson/runs/10432720183

patricoferris:
https://ci.ocamllabs.io/github/geocaml/ocaml-topojson/commit/6b834a2b17e36cf98a826d3b42c49315f1288f21/variant/debian-11-5.0_arm64_opam-2.1

patricoferris:
Thanks! ^^

PR #39, opened 2023-01-16
Map_object  ---- (Implement a non-blocking, streaming codec for TopoJSON)

Priceless-P:


Priceless-P:
Hi @patricoferris Please is the `map_object` doing what it ought to?

src/topojsonm/topojsonm.ml
```diff
@@ -1 +1,227 @@
...
+    | `Lexeme _ as t ->
+        enc t;
+        go ()
+    | `Error e -> raise (Abort (`Error (loc (), e)))
+    | `End -> ignore @@ Jsonm.encode encoder `End
+    | `Await -> assert false
+  in
+  try Ok (go ()) with Abort e -> Error e
+
+let map_props f src dst =
```
patricoferris: Let's remove any imported code from Geojson that we don't need, otherwise the PR is quite big and hard to review.

src/topojsonm/topojsonm.ml
```diff
@@ -1 +1,227 @@
...
+  let encoder = Jsonm.encoder dst in
+  let loc () = Jsonm.decoded_range decoder in
+  let enc v =
+    match Jsonm.encode encoder v with
+    | `Ok -> ()
+    | `Partial -> raise (Abort (`Unexpected "partial encoding"))
+  in
+  let rec go () =
+    match Jsonm.decode decoder with
+    | `Lexeme (`Name "object" as t) ->
```
patricoferris:
```suggestion
    | `Lexeme (`Name "objects" as t) ->
```

src/topojsonm/topojsonm.ml
```diff
@@ -1 +1,227 @@
...
+        (match Jsonm.decode decoder with
+        | `Lexeme (`Name "example" as ex) -> (
```
patricoferris: We don't want to _only_ pull out one particular case in the `objects` field, we want to iterate over all of them providing the function `f` with both the name associated with the object _and_ the decoded object converted to the Topojson value.

src/topojsonm/topojsonm.mli
```diff
@@ -1 +1,115 @@
...
+  (unit, Err.t) result
+(** [map_geometry f src dst] will apply [f] to all TopoJson objects. This is
+    essentially any
+    {{:https://datatracker.ietf.org/doc/html/rfc7946#section-3.1} geometry
+    object}.
+
+    The map will recurse into geometry collections. Note for the moment if you
+    have a single geometry object as your document, this will not work. *)
+
+val map_props :
```
patricoferris: Again let's remove anything unnecessary for this PR.

src/topojsonm/vendor/ezjsonm/ezjsonm.ml
```diff
@@ -0,0 +1,384 @@
+(*
```
patricoferris: If we make a dependency on `Geojsone` then we can just use the module exported there rather that re-writing all of this code.

patricoferris (commented):
Getting there :))

src/topojsonm/vendor/ezjsonm/ezjsonm.ml
Priceless-P: Okay. How do we make a dependency on `Geojsone`? By including it in `dune` file?

src/topojsonm/topojsonm.mli
Priceless-P: Okay so for clarity's sake. The only functions we need here are 

- `map_object`
- `fold_object`
- `iter_object`

Yes?

patricoferris: Let's just start with `map_object`

src/topojsonm/vendor/ezjsonm/ezjsonm.ml
patricoferris: Yep, then the `topojsonm.opam` file

Priceless-P:
@patricoferris  I'm not sure about this 
```ocaml
let b = Jsonm.decode decoder in
```
What do you think?

And is the `topojsonm.opam` okay?

patricoferris:
Ah the opam file is generated from the `dune-project` file sorry I should have mentioned that

src/topojsonm/topojsonm.ml
```diff
@@ -171,16 +171,17 @@ let map_object f src dst =
...
+        | `Os -> (
+        let b = Jsonm.decode decoder in
             match Topo.of_json @@ decode_single_object decoder with
             | Error (`Msg m) -> raise (Abort (`Unexpected m))
             | Ok v -> (
                 match Topo.topojson v with
                 | Geometry g ->
-                    let g' = f g in
-                    enc (`Lexeme ex);
+                    let g' = f b g in
```
patricoferris: This isn't quite right, remember as well we want to pass the name and geometry to `f`. Here's a function that you might find useful. 

```ocaml
let rec loop_through_objects decoder = match Jsonm.decode decoder with
    | `Lexeme `Oe -> ignore (enc (`Lexeme `Oe)); ()
    | `Lexeme (`Name geometry_name) -> (
      let geometry_json = decode_single_object decoder in
      match Topo.Geometry.of_json geometry_json with
      | Error (`Msg m) -> failwith m
      | Ok g ->
        let new_name, new_geometry = f (geometry_name, g) in
        enc (`Lexeme (`Name new_name));
        encode_value encoder (Topo.Geometry.to_json new_geometry);
        loop_through_objects decoder
    )
    | _ -> failwith "Unexpected lexeme"
  in
```

This function allows you to loop through the objects, apply the function and re-encode them. That combined with all of the work in this PR should get you pretty close to a working solution.

src/topojsonm/topojsonm.ml
```diff
@@ -171,16 +171,17 @@ let map_object f src dst =
   in
   let rec go () =
     match Jsonm.decode decoder with
-    | `Lexeme (`Name "object" as t) ->
+    | `Lexeme (`Name "objects" as t) ->
         (match Jsonm.decode decoder with
-        | `Lexeme (`Name "example" as ex) -> (
+        | `Os -> (
```
patricoferris: Remember we're also building the output as we go, so you will need to encode these values. 

Unfortunately the API was built with polymorphic variants so there's no error here but `Jsonm.decode` cannot return just `` `Os`` it has to be `` `Lexeme `Os``

Priceless-P:
How about now ?

src/topojsonm/topojsonm.ml
```diff
@@ -171,27 +171,33 @@ let map_object f src dst =
...
-                    let g' = f b g in
-                    enc (`Lexeme t);
-                    encode_value encoder
-                      (Topo.to_json @@ Topo.v ?bbox:(Topo.bbox v) (Geometry g'));
-                    go ()
-                | _ -> raise (Invalid_argument "Expected a geometry object")))
+    | `Lexeme (`Name "objects" as t) -> (
+        match Jsonm.decode decoder with
+        | `Lexeme `Os ->
+            let rec loop_through_objects decoder =
```
patricoferris: I wouldn't define it inside here but move it completely outside of this function.

src/topojsonm/topojsonm.ml
```diff
@@ -171,27 +171,33 @@ let map_object f src dst =
...
+                  match Topo.Geometry.of_json geometry_json with
+                  | Error (`Msg m) -> failwith m
+                  | Ok g ->
+                      let new_name, new_geometry = f (geometry_name, g) in
+                      enc (`Lexeme (`Name new_name));
+                      encode_value encoder (Topo.Geometry.to_json new_geometry);
+                      loop_through_objects decoder)
+              | _ -> failwith "Unexpected lexeme"
+            in
+            enc (`Lexeme t);
```
patricoferris: This probably needs to come up near where `t` is defined otherwise we're encoding things after things have been encoded!

src/topojsonm/topojsonm.ml
```diff
@@ -171,27 +171,33 @@ let map_object f src dst =
...
-                | Geometry g ->
-                    let g' = f b g in
-                    enc (`Lexeme t);
-                    encode_value encoder
-                      (Topo.to_json @@ Topo.v ?bbox:(Topo.bbox v) (Geometry g'));
-                    go ()
-                | _ -> raise (Invalid_argument "Expected a geometry object")))
+    | `Lexeme (`Name "objects" as t) -> (
+        match Jsonm.decode decoder with
+        | `Lexeme `Os ->
```
patricoferris, line 193: Doesn't look like this lexeme gets encoded

patricoferris (commented):
I would write a test that uses the `map_objects` and does something to the name or geometry so you can check things are working properly :))

test/topojsonm/expected/test.ml
```diff
@@ -15,19 +15,20 @@ let test_map_objects_topojson () =
...
       | _ -> geometry
     in
     (new_name, new_geometry)
   in
   let res = Topojsonm.map_object f src dst in
+
   close_in file;
   match res with
   | Ok () ->
       let buff = Buffer.create 1000 in
```
patricoferris: This buffer is empty! You need to get the contents of the buffer you passed as the encoder and then it should work :))

Priceless-P: Will this do it ?
```ocaml
let file = open_in "../../topojson/test_cases/files/exemplar.json" in
...
let buff = Buffer.create 1000 in
      (try Buffer.add_channel buff file max_int with _ -> ());
      let json_str = Buffer.contents buff in
print_string json_str
```

patricoferris: The buffer is already defined at the top

https://github.com/geocaml/ocaml-topojson/blob/7e61172ed9a351c7fef7f6fa13ebdb53b2f848dc/test/topojsonm/expected/test.ml#L6

Priceless-P: The `d` is of type `jsonm.encoder` but I need a `Buffer.t`

I tried to do
```ocal
let json_str = Buffer.contents d.Jsonm.dst
```
But I get `Unbound record field Jsonm.dst`

patricoferris: You just need to split the two:

```ocaml
 let buff = Buffer.create 1024 in
 let d = Jsonm.encoder (`Buffer buffer) in 
```

Priceless-P: Ohh
Thank you very much sir.

Priceless-P: I noticed the `example` is being modified to `new_example` as expected. 
However, I don't see any modification in the geometry object itself. 
What did I miss? 🤔

patricoferris: The entry you are trying to change is actually inside a GeometryCollection not a top-level object, so all of objects are going through the `_ -> geometry` case.

Priceless-P: Okay. I have changed that. It seems to be working now. What do you think?

patricoferris:
Awesome! You just need to promote the results to that file with a `dune promote` and we should be good. This is excellent work!! :)))

Priceless-P:
Thank you  :))

I'm learning from the best teacher ever 🎉

`dune promote` okay now?

patricoferris:
Thank you ^^ The tests are greeeeen :)

PR #41, opened 2023-01-21
fold_object ---- (Implement a non-blocking, streaming codec for TopoJSON)

Priceless-P:
`fold_object` is like map_object but allows you to accumulate some result that is then returned to you.


Priceless-P:
Hi @patricoferris, What do you think? The test fails sometimes.

src/topojsonm/topojsonm.mli
```diff
@@ -37,13 +37,17 @@ val map_object :
...
-    object}.
-
-    The map will recurse into geometry collections. Note for the moment if you
-    have a single geometry object as your document, this will not work. *)
+(** [map_object f src dst] will apply [f] to all TopoJson objects. The map will
+    recurse into TopoJson Object. Note for the moment if you have a single
+    geometry object as your document, this will not work. *)
+
+val fold_object :
+  ('a -> string * Topo.Geometry.t -> 'a) ->
```
patricoferris: Shall we call this acc instead to make it clearer what it is?

```suggestion
  ('acc -> string * Topo.Geometry.t -> 'acc) ->
```

test/topojsonm/expected/test.ml
```diff
@@ -1,11 +1,12 @@
 (* Tests to validate implemented functions/ modules  *)
+let file = open_in "../../topojson/test_cases/files/exemplar.json"
```
patricoferris: Let's keep this inside a function call

patricoferris (commented):
This looks really good thank you! ~We just need to make sure we are still testing the map function too. We can call both and have everything go to the test.expected file~ It does, ignore me!

test/topojsonm/expected/test.ml
Priceless-P: Okay. Sure

Priceless-P: Done.

Priceless-P:
I am guessing the test is failing because we are not encoding the `transform` object (yet?)

patricoferris:
It was failing because of a bug! We weren't decoding/encoding bounding boxes correctly, see the fix (which has now been merged) in #42 -- if you rebase your branch on `main` it should be good to go

Priceless-P:
Done. But the test is still failing on my end 😥

patricoferris:
Another bug ^^" very similar to the last one, thank you for finding them! Should be good to rebase again and try

Priceless-P:
Works perfectly now :))

test/topojsonm/expected/test.ml
```diff
@@ -34,4 +33,22 @@ let test_map_objects_topojson () =
       Topojsonm.Err.pp Format.err_formatter e;
       failwith "Internal err"
 
-let () = test_map_objects_topojson ()
```
patricoferris, line 37: We need to still call the `test_map_objects_topojson` function in the main `let () =` -- that's why the tests are failing in CI I think at the moment

patricoferris:
And promote the new result of mapping because you modified the test case :))

patricoferris:
Thanks!!!

Priceless-P:
💃🏻💃🏻💃🏻💃🏻

PR #45, opened 2023-01-28
topojsone docs

Priceless-P:
`Topojsone` functions documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ But first we must confirm that it is a topology.
 ```
 
 # Topojsone
-Topojson is a non-blocking, streaming parser for TopoJSON objects. It uses jsone and [Eio](https://github.com/ocaml-multicore/eio) library.
+Topojson is a non-blocking, streaming parser for TopoJSON objects. It uses jsone and it is up to the user to provide non-blocking versions of the source and destination functions for the parser to be truly non-blocking. One such library that can provide these functions is [Eio](https://github.com/ocaml-multicore/eio) and that is what we shall use for the rest of the example.
 
 ## Sources
 `src_of_flow` takes a "flow" and returns a function that reads from the flow into a buffer and returns a substring of the buffer.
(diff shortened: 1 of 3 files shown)


Priceless-P:
Hi @patricoferris, please review

README.md
```diff
@@ -125,3 +130,103 @@ But first we must confirm that it is a topology.
   | Topology t -> Topology.foreign_members t;;
 - : (string * json) list = [("extra", `A [`String "Wow!"])]
 ```
+
+# Topojsone
+Topojson is a non-blocking, streaming parser for TopoJSON objects. It uses jsone and [Eio](https://github.com/ocaml-multicore/eio) library.
```
patricoferris: We don't actually depend on Eio (except in the tests), you'll notice our source and destinations are actually just functions that fill and read buffers, it is up to the user to provide non-blocking versions of these functions in order for the parser to be truly non-blocking. I would maybe state that and then say one such library that can provide these functions is Eio and that is what we shall use for the rest of the example :))

README.md
```diff
@@ -125,3 +130,103 @@ But first we must confirm that it is a topology.
...
+
+## Sources
+`src_of_flow` takes a "flow" and returns a function that reads from the flow into a buffer and returns a substring of the buffer.
+```ocaml
+# let src_of_flow flow =
+  let buff = Cstruct.create 2048 in
+  fun () ->
+    let got = Eio.Flow.(read flow buff) in
```
patricoferris:
```suggestion
    let got = Eio.Flow.(single_read flow buff) in
```

README.md
```diff
@@ -125,3 +130,103 @@ But first we must confirm that it is a topology.
...
+    let t = Cstruct.sub buff 0 got in
+    t;;
+Line 4, characters 25-29:
+Alert deprecated: Eio.Flow.read
+Use single_read if you really want this.
+val src_of_flow : #Eio.Flow.source -> unit -> Cstruct.t = <fun>
+```
+
+`with_src` opens a file with a given name, creates a source of flow using the file, then calls a given function with that source of flow as an argument.
```
patricoferris:
```suggestion
`with_src` opens a file with a given name creating a flow in the process. We can then use `source_of_flow` to turn this into a function that is compatible with the `Topojsone` API.
```

README.md
```diff
@@ -125,3 +130,103 @@ But first we must confirm that it is a topology.
...
+```ocaml
+# let with_src cwd f func =
+  Eio.Path.(with_open_in (cwd / f)) @@ fun ic -> func @@ src_of_flow ic;;
+val with_src :
+  #Eio.Fs.dir Eio.Path.t -> string -> ((unit -> Cstruct.t) -> 'a) -> 'a =
+  <fun>
+```
+
+## Destinations
+`buffer_to_dst` is a function that takes a buffer and a Cstruct, and turns it into a destination by coping the contents of the Cstruct into the buffer using the Eio.Flow.copy function.
```
patricoferris:
```suggestion
Destinations are functions that take a buffer full of data and should put that somewhere. `buffer_to_dst` is a function that takes a buffer and a Cstruct, and turns it into a destination by copying the contents of the Cstruct into the buffer using the `Eio.Flow.copy` function.
```

patricoferris (commented):
Excellent! I've added a few clarifying comments but other than that it's really good, thank you.

README.md
Priceless-P: Okay.😊

src/topojsonm/dune
```diff
@@ -1,4 +1,4 @@
 (library
  (name topojsone)
  (public_name topojsone)
- (libraries topojson geojsone))
+ (libraries topojson eio geojsone))
```
patricoferris:
```suggestion
 (libraries topojson geojsone))
```

There should be no dependency on Eio for the library.

patricoferris:
Great thank you!!!

PR #46, opened 2023-02-04
extract functions (Geojson to Topojson)

Priceless-P:
Extracts lines and rings from Geojson's Objects


Priceless-P:
@patricoferris what do i think about this? I also intend to add `extract_geomteries` and `extract_features_collection` (if necessary).

src/topojson/topojson.ml
```diff
@@ -616,4 +616,54 @@ module Make (J : Intf.Json) = struct
...
+                match hd with
+                | `List c -> List.map J.to_float c
+                | _ -> failwith "Invalid coordinate format"
+              in
+              loop (coord :: acc) tl
+        in
+        loop [] coordinates
+    | _ -> failwith "Not a valid LineString or Polygon"
+
+  let extract_lines = function
```
patricoferris: This comment applies to pretty much most of the other functions defined here.

I don't think we want to reimplement all of the JSON parsing logic again, GeoJSON does this already for us so we should just uses GeoJSON to extract the coordinates etc. from the geometries (since we are converting from GeoJSON to TopoJSON).

src/topojson/topojson_intf.ml
```diff
@@ -282,6 +282,21 @@ module type S = sig
 
     val to_json : ?bbox:float array -> t -> json
     val of_json : json -> (t, [ `Msg of string ]) result
+
+    val extract_lines :
```
patricoferris: We won't actually ever expose these internal functions. Instead, eventually, we will just have something like:

```ocaml
val geojson_to_topojson : Geojson.t -> Topojson.t
```

src/topojson/topojson.ml
patricoferris: e.g. 

https://github.com/geocaml/ocaml-geojson/blob/d45db9166a050da2e61ff3f04856c4788c13edc8/src/geojson/geojson_intf.ml#L124

Priceless-P:
Hi @patricoferris, 
```ocaml
module Geojson = Geojson.Make (Ezjsonm_parser)
```
keep giving error `Unbound module Ezjsonm_parser`

Also with the current implementation, will `extract_lines_from_geometry` also extract rings from `Polgon` and `MultiPolygon`?

patricoferris:
> Also with the current implementation, will extract_lines_from_geometry also extract rings from Polgon and MultiPolygon?

I think so, they are after all just lines. 

> keep giving error Unbound module Ezjsonm_parser

Yep, we still want to be agnostic to the JSON parser the user decided to use. You should be able to pass the same one that gets passed to our `Make` "functor". This is a special name for modules that are _parameterised_ by other modules.

```ocaml
module Make (J : Intf.Json) = struct
  module Geojson = Geojson.Make(J)
  ...
```

Priceless-P:
Hi @patricoferris what do you think?

src/topojson/topojson.ml
```diff
@@ -616,50 +616,74 @@ module Make (J : Intf.Json) = struct
...
+    | Geometry g -> (
+        let geo = geometry g in
+        match geo with
+        | LineString ls -> [ LineString.coordinates ls ]
+        | MultiLineString ml ->
+            let lines = MultiLineString.lines ml in
+            let arr = Array.map LineString.coordinates lines in
+            Array.to_list arr
+        | Polygon p ->
+            let rings = Polygon.rings p in
```
patricoferris, line 636: Hmm thinking about this a little more, perhaps we have to treat rings slightly differently because the first and last point are technically connected.

src/topojson/topojson.ml
```diff
@@ -616,50 +616,74 @@ module Make (J : Intf.Json) = struct
...
+      Geojson.Geometry.Position.t list =
+    let open Geojson.Geometry in
+    let junction_table : (Position.t, Position.t list) Hashtbl.t =
+      Hashtbl.create (List.length lines)
+    in
+    let junction : Position.t list = [] in
+    List.iter
+      (fun line ->
+        let line_length = Array.length line in
+        for i = 0 to line_length - 2 do
```
patricoferris: Points on a line other than the first and the last have two neighbours: `n - 1`  and `n + 1`.

src/topojson/topojson.ml
```diff
@@ -616,50 +616,74 @@ module Make (J : Intf.Json) = struct
...
+  let extract_from_feature_collection (fc : Geojson.Feature.Collection.t) :
+      Geojson.Geometry.Position.t array list =
+    let open Geojson.Feature.Collection in
+    let features = features fc in
+    List.concat (List.map extract_from_feature features)
+
+  let find_junctions (lines : Geojson.Geometry.Position.t array list) :
+      Geojson.Geometry.Position.t list =
+    let open Geojson.Geometry in
+    let junction_table : (Position.t, Position.t list) Hashtbl.t =
```
patricoferris: Why not make this a position to position set hash table to stay closer to how the blog post does it?

src/topojson/topojson.ml
```diff
@@ -616,50 +616,74 @@ module Make (J : Intf.Json) = struct
...
+      Hashtbl.create (List.length lines)
+    in
+    let junction : Position.t list = [] in
+    List.iter
+      (fun line ->
+        let line_length = Array.length line in
+        for i = 0 to line_length - 2 do
+          let point1 = line.(i) in
+          let point2 = line.(i + 1) in
+          if Hashtbl.mem junction_table point1 then
```
patricoferris, line 707: I think for every point we visit we want to put in the hash table it's neighbours as a set. So if we are at point `N` then we but `N, { N - 1, N - 2}` as a key-value pair into the hash table.

Then if we come across `N` again we check that `N`s neighbours are different (therefore a junction) or if they are the same then it isn't.

Priceless-P:
@patricoferris I am having a lot of type incompatibility errors. Please help me out 😥

patricoferris:
Ah, most of the issues were just related to giving functions a little too much generality and also we didn't expose that the positions were float arrays. Feel free to cherry-pick this commit https://github.com/geocaml/ocaml-topojson/commit/7b31814b8933c11284343203e9e97122f6baf092

Priceless-P:
The type errors are gone now. :)) Thank you @patricoferris

Priceless-P:
Do you think the `find_junction` is correction? I changed the MultiLineString coordinates to 
```json
 "coordinates": [
        [
			[1, 1], [2, 2],[3, 3],
			[4, 4], [3, 1],[4, 2]

        ], [
			[4, 2], [4, 4], [4, 6]
        ]
    ]
```
And the test didn't print out any junctions

patricoferris:
I don't think the function is correct, see my comments inline on the code.

test/topojson/test.ml
```diff
@@ -127,7 +127,7 @@ let pp_transform ppf t =
 
 let transform = Alcotest.testable pp_transform Stdlib.( = )
 
-let to_string pp a =
+let _to_string pp a =
```
patricoferris, line 130: If this function isn't being used we can probably remove it ?

src/topojson/topojson.ml
```diff
@@ -698,24 +696,23 @@ module Make (J : Intf.Json) = struct
...
+    let junction : Position.t list ref = ref [] in
     List.iter
       (fun line ->
         let line_length = Array.length line in
-        for i = 1 to line_length - 2 do
+        for i = 0 to line_length - 1 do
           let point1 = line.(i) in
-          let point2 = line.(i - 1) in
-          let point3 = line.(i + 1) in
+          let point2 = if i > 0 then line.(i - 1) else line.(line_length - 1) in
```
patricoferris, line 705: Isn't this only true if we're talking about a ring rather than any line?

src/topojson/topojson.ml
```diff
@@ -698,24 +696,23 @@ module Make (J : Intf.Json) = struct
...
+          let point3 = if i < line_length - 1 then line.(i + 1) else line.(0) in
```
patricoferris, line 706: Same question here?

src/topojson/topojson.ml
```diff
@@ -698,24 +696,23 @@ module Make (J : Intf.Json) = struct
     let junction_table : (Position.t, Set.t) Hashtbl.t =
```
patricoferris, line 696: We really want to know _every_ set of neighbours to begin with (let's get the algorithm correct before optimising). So I think we want something like: `(Position.t, Set.t list) Hashtbl.t`.

src/topojson/topojson.ml
```diff
@@ -698,24 +696,23 @@ module Make (J : Intf.Json) = struct
...
-        for i = 1 to line_length - 2 do
+        for i = 0 to line_length - 1 do
           let point1 = line.(i) in
-          let point2 = line.(i - 1) in
-          let point3 = line.(i + 1) in
+          let point2 = if i > 0 then line.(i - 1) else line.(line_length - 1) in
+          let point3 = if i < line_length - 1 then line.(i + 1) else line.(0) in
           if Hashtbl.mem junction_table point1 then
             let neighbors = Hashtbl.find junction_table point1 in
             Hashtbl.replace junction_table point1
```
patricoferris, line 709: Instead of adding the neighbours to an ever growing set, I think we want to keep them distinct so we know what two neighbours belong together. Consider the problem of having two lines `ABC` and `CBA`. `B` in this case is _not_ a junction because the neighbours `A` and `C` are the same in both lines.

src/topojson/topojson.ml
```diff
@@ -698,24 +696,23 @@ module Make (J : Intf.Json) = struct
...
               (Set.add point2 (Set.add point3 neighbors))
           else
             Hashtbl.add junction_table point1 (Set.of_list [ point2; point3 ]);
           let neighbors = Hashtbl.find junction_table point1 in
-          if Set.cardinal neighbors > 2 then
-            ref junction := point1 :: !(ref junction)
+          if Set.cardinal neighbors > 2 then junction := point1 :: !junction
```
patricoferris, line 714: With the above changes this will become `List.length` when we store the list of possible neighbours and store a new pair _iff_ they are different as defined in the algorithm.

patricoferris (commented):
These changes are heading in the right direction, nice! Comments are inline.

